PR #23, opened 2018-05-24
pps: remove input_file_id from get_logs signature

williballenthin:
in at least pach 1.7.3, there is no `input_file_id` field in the `GetLogsRequest`, so asking for logs raises an exception.

this fixes the issue on pach 1.7.3, but obviously removes support for some older versions (such as 1.6.5). i dont know what your strategy for dealing with versions is.


David-Development:
Any updates on this one? Ran into this issue as well.. Great to see that someone fixed it already!

ysimonson:
Thanks for this fix!

PR #34, opened 2019-03-05
De-bitrot

ysimonson:
* Remove flaky `make test` target in favor of tox everywhere
* Update protobufs to pachyderm v1.8.5
* Switch to python v3.7
* Fix outdated test
* Pin protobuf compilation dependencies
* Fixed a bug where tox wasn't updating deps
* Tweaks to readme


David-Development:
Awesome! :+1: 

The `get_logs` method in the `pps_client.py` does not support filtering by a given datum. The following works for me:

```python
def get_logs(self, pipeline_name=None, job_id=None, data_filters=tuple(), master=False, datum=None):
    pipeline = Pipeline(name=pipeline_name) if pipeline_name else None
    job = Job(id=job_id) if job_id else None
    if pipeline is None and job is None:
        raise ValueError("One of 'pipeline_name' or 'job_id' must be specified")
    return list(self.stub.GetLogs(
        GetLogsRequest(pipeline=pipeline, job=job, data_filters=data_filters, master=master, datum=datum)))
```


Also when running `pps_client.list_job()`, the `input` and `stats` field will be empty.

ysimonson:
@David-Development thanks! This PR will just get the python lib back to a working state, but will take a look at what parts of the API need modernizing after that, including `get_logs`.

ysimonson:
@David-Development if you open a PR with those changes I'd be happy to accept them! Otherwise I can tackle it at a later point.

David-Development:
@ysimonson 
> Also when running pps_client.list_job(), the input and stats field will be empty.

This is already fixed in the latest version 🎉 

The `get_logs` part is not working. As you already merged #23, adding `datum=None` to the method signature and adding it to the actual method call should be sufficient in order to get that feature working.

One thing that I noticed though.. When calling `get_logs`, it'll just say something like "pod doesn't exist" when using `standby: true` (same behavior in pachctl). It would be cool if the pod doesn't exist anymore and the user enabled stats, that the logs will be pulled out of the stats branch. What do you think about it?

ysimonson:
I think it makes sense to keep this library semantically the same as what the raw gRPC calls would give you, with the few divergences that exist to strictly make the interface more pythonic - so, IMO, that'd make more sense as a utility function within individual projects rather than part of the library.

David-Development:
@ysimonson makes sense. Thought that this might be a feature that would help quite a few users as it requires a bit of programming.

PR #41, opened 2019-04-03
adds function for getting the remote version of pachyderm Closes #37

### LINKED ISSUE #37: Feature request: Add get_version()
It would be great to have an equivalent for `pachctl version` in this package, so we can retrieve the version of pachd we're talking to.

christianladron:


ysimonson:
Thanks @christianladron! TIL `contextlib.closing`, that's pretty cool. Would you mind adding a test for this?

tests/test_get_version.py
```diff
@@ -0,0 +1,16 @@
...
+"""Tests for the `get_remote_version` function of the `python_pachyderm` package."""
+import python_pachyderm
+
+def test_pfs_client_init_with_default_host_port():
+    # GIVEN a Pachyderm deployment
+    # WHEN a client requests the remote version of a pachyderm deployment
+    version_pb = python_pachyderm.get_remote_version()
+    remote_version = "{}.{}.{}\n".format(
+            version_pb.major, version_pb.minor, version_pb.micro)
+    # THEN remote version should match the pachctl version which deployed it
```
ysimonson: This might not be true; you can have valid setups where pachctl and the server versions are different. I think it's easier to just verify that you got a response back from the server when you requested the version, rather than match it against the VERSION file. This has the added benefit that you don't need to copy the VERSION file around on CI setup, which means tests won't break when you run them locally.

christianladron: Ok, I made this thinking only in the Travis test case, where the remote and local version match.

PR #48, opened 2019-04-17
list_job input_commit argument is a list now, and output_commit argum…

christianladron:
…ent added


ysimonson:
This makes sense, since it's more in-line with the proto service. But could you make the `input_commit` argument backwards compatible with any existing use of this function? Something like this:

```python
if isinstance(input_commit, list):
    input_commit = [commit_from(ic) for ic in input_commit]
else:
    input_commit = commit_from(input_commit)
```

ysimonson:
Thanks!

PR #55, opened 2019-04-30
pps_client issues

jfri3d:
Minimal changes to three functions to mimic functionality in `pachctl`:
- `input_commit`
   - removed dependency from `list_job`
- `inspect_datum()`
  - swapped `proto.Datum()` argument to `datum_id` and `job_id`
- `delete_pipeline()`
  - removed "reserved" args given errors

### DIFF
--- a/src/python_pachyderm/pps_client.py
+++ b/src/python_pachyderm/pps_client.py
@@ -2,6 +2,8 @@
 
 from __future__ import absolute_import
 
+import six
+
 from python_pachyderm.client.pps import pps_pb2 as proto
 from python_pachyderm.client.pps import pps_pb2_grpc as grpc
 from python_pachyderm.util import commit_from, get_address, get_metadata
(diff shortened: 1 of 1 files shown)


ysimonson:
The features being removed here are generally exposed as flags in `pachctl`. Moreover, this would break backwards compatibility, which I'd like to minimize even though it's occasionally necessary.

Is there something specific you're hoping to achieve? Maybe there's an alternate route.

jfri3d:
Hi Yusuf. Apologies for the *overzealous* "fixes". I was unable to call `client.list_job()` and `client.delete_pipeline(<name>)`.

ysimonson:
Ah, that makes sense. Let me review the rest of the changes and we can get this merged in.

src/python_pachyderm/pps_client.py
```diff
@@ -45,11 +43,11 @@ def inspect_job(self, job_id, block_state=False):
     def list_job(self, pipeline=None, input_commit=None, output_commit=None):
         if isinstance(input_commit, list):
             input_commit = [commit_from(ic) for ic in input_commit]
-        else:
+        elif isinstance(input_commit, str):
             input_commit = [commit_from(input_commit)]
         if output_commit:
             output_commit = commit_from(output_commit)
-        req = proto.ListJobRequest(pipeline=pipeline, input_commit=input_commit, output_commit=output_commit)
+        req = proto.ListJobRequest(pipeline=proto.Pipeline(name=pipeline), input_commit=input_commit, output_commit=output_commit)
```
ysimonson: Since name is the only field for pipelines, this makes sense. But could you rename the field to `pipeline_name`? That'll be more consistent with naming conventions elsewhere, and reduce the chances of type errors from codebases using the old function signature.

ysimonson:
Thanks!

PR #63, opened 2019-05-23
62: exposing history option in list_file

### LINKED ISSUE #62: missing history option
**Motivation**
The default method for listing files in a repo "merges" all commits together (shown below). The *true* commit history is only available with the `--history` option, which is not exposed in the python client. 

**Goal**
Expose the `--history` option from the python client.

image

jfri3d:
Closes #62 .


ysimonson:
Looks good, thank you!

PR #123, opened 2019-07-22
Adds TLS support

christianladron:
Adds keyword arument root_certificates to PpsClient and PfsClient to
create a secure client as described in https://grpc.github.io/grpc/python/_modules/grpc.html#secure_channel


ysimonson:
@christianladron thanks for this! Looks like there's a couple of small flake8 formatting issues remaining.

You can check flake8 locally (rather than waiting for the whole TravisCI test cycle to complete) via `flake8 src/python_pachyderm --exclude=src/python_pachyderm/client --max-line-length=120 --max-doc-length=79  --ignore=F401`

PR #124, opened 2019-08-08
Fix docs lists

jashandeep-sohi:
Reading the `pdoc` rendered docs is a little hard because the params list don't render correctly. This should fix that.


ysimonson:
Thanks for this! Looks like there's a couple of formatting errors with this change, could you fix them?

jashandeep-sohi:
oh yeah, fixed

PR #128, opened 2019-09-17
Reorganize - First Variant

ysimonson:
This is a large reorganization of the module layout:

* Renamed `python_pachyderm.client` renamed to `python_pachyderm._proto`.
* Renamed `pps_client` and `pfs_client` to `pps` and `pfs`, respectively.
* Introduced a `version` package, and moved `get_remote_version` there.
* Raw protos are no longer exported at the top-level module.

Benefits:
* It provides clearer naming.
* Manual mappings of proto to python names would no longer need to be maintained. This would make ongoing maintenance easier.

Drawbacks:
* PFS, PPS and version protos are star imported, triggering some warnings that need to be disabled, and including attributes that library users probably shouldn't use.
* It's backwards incompatible, and would require some tweaks from most library users.


ysimonson:
Closing in favor of #129, see discussion there for an explanation why.

PR #136, opened 2019-10-23
[WIP] Run tests against multiple pachyderm versions

ysimonson:
This is a WIP effort to run the test matrix against multiple pachyderm core versions


ysimonson:
Test runs are proving that the tests would need to have a substantial number of new conditionals based off of the pachyderm version. This is doable, but would require a lot more maintenance effort, so closing this out for now.

PR #143, opened 2019-10-29
Initialization options

### LINKED ISSUE #140: Drop support for `PACHD_ADDRESS`
This library currently supports the `PACHD_ADDRESS` as a way of configuring the `Client`. However, `PACHD_ADDRESS` is [marked for deprecation](https://github.com/pachyderm/pachyderm/issues/4171).

### LINKED ISSUE #141: Add first-class support for in-cluster `Client` instances
Similar to the go client, we should add support for instantiating a `Client` when in-cluster

ysimonson:
* Removed support for `PACHD_ADDRESS` (closes #140)
* Added first-class support for in-cluster clients (closes #141)


ysimonson:
@msteffen Kind of curious to get your thoughts here, in particular the addition of pachd address parsing. The downside is that this is copying the same address parsing logic that's been added to the go client, and the original intent was to avoid that. But on the other hand, being able to just copy in config values or pachd address values pulled from elsewhere seems convenient.

PR #152, opened 2019-11-01
header_records option of PutFileRequest class

amoallim15:
header_records option of PutFileRequest class is missing!


ysimonson:
Good catch, thank you! I'll merge this shortly, but first want to check if some improvements to the proto linter will allow it to catch missing features like this in the future.

PR #167, opened 2020-01-10
Support for v1.9.10

ysimonson:


ysimonson:
Looks like, while `BatchTransaction` is in master, it wasn't part of v1.9.10. Rather than do a separate release for both 1.9.10 and 1.9.11, I'll just wait until 1.9.11 is out and re-open this PR.

PR #168, opened 2020-01-14
Better docs

ysimonson:
We currently rely on [pdoc](https://github.com/mitmproxy/pdoc), which has a number of bugs and appears to be unmaintained. This rebuilds the docs with [pdoc3](https://pdoc3.github.io/pdoc/), a fork that appears to fix most of the issues.


ysimonson:
While these docs make some improvements and allow us to move off of a custom fork of pdoc, it introduces a few problems of its own:

* Aliased imports are redundantly documented, making the docs much larger than they need to be
* ...Yet, class members from mixins aren't redundantly documented (though they link off to the mixin docs), making `Client` less comprehensible (especially to people who don't know `Client`'s functionality basically all comes from mixins.)
* Enum members are still not documented.

In the long-run, it'll probably be the right move to switch to pdoc3; however, it's not necessary now as our fork of pdoc still works. Maybe when we do need to switch to pdoc3, a some or all of these issues will be addressed. In the mean-time, closing this out.

PR #203, opened 2020-07-24
Added version number to build

### LINKED ISSUE #202: Add version number dunder to package
I find it useful to have the version number dunder to check against when I get odd behavior. 

```
> python_pachyderm.__version__
AttributeError                            Traceback (most recent call last)
<ipython-input-28-7a2d32cfa4e4> in <module>
----> 1 python_pachyderm.__version__

AttributeError: module 'python_pachyderm' has no attribute '__version__'
```

JimmyWhitaker:
Fixes #202

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ __pycache__/
 
 # VSCode
 .vscode
+
+# python-pachyderm
+src/python_pachyderm/version.py
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,9 @@ with open(join(dirname(__file__), "version.json"), "r") as f:
     j = json.load(f)
     version = j["python-pachyderm"]
 
+with open(join(dirname(__file__), 'src', 'python_pachyderm','version.py'), 'w') as f:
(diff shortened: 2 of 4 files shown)


ysimonson:
Would you know if this is the usual way python libraries set `__version__`? Or do they hard-code it? Or something else?

JimmyWhitaker:
So, it seems as though there were no less than a dozen ways to do it. Most
of which required hard coding it in the init file of the repository.  This
way seemed to be the best way to make it dependent on the current
versioning scheme, allowing it to be dependent on the build mechanism
rather than manual edits of a python file. I stole this approach from
pytorch, so I assumed it was thought through to some extent. Happy to do it
another way if you prefer.

ysimonson:
Cool, make sense! Just a couple of things in that case:

* Lint is currently failing, could you fix? You can run `make lint` locally to verify.
* Could you add a test for this? Usually it wouldn't matter for something as small as an attribute, but since this is dynamically generated I think it's worth testing.

JimmyWhitaker:
Sure thing. I'll look into it and get back to you if there are any issues.

PR #214, opened 2020-08-11
Update spout docstring

Tryneus:
This docstring was missed in #206


ysimonson:
Could you merge or rebase master in? Should fix CI.

ysimonson:
I just realized this isn't quite right - it's using `SpoutManager` itself as a context manager still. I'll fix this + CI in a new PR.

PR #225, opened 2020-09-14
Make python spout manager tolerant of exceptions

msteffen:
Just creating this PR in preparation for the oncall handoff. We probably need a test for this as well, plus a new release

### DIFF
--- a/src/python_pachyderm/spout.py
+++ b/src/python_pachyderm/spout.py
@@ -64,7 +64,7 @@ class SpoutManager:
             yield spout_commit
         finally:
             spout_commit.close()
-        self._has_open_commit = False
+            self._has_open_commit = False
 
 
 class SpoutCommit:
--- a/src/python_pachyderm/spout.py
+++ b/src/python_pachyderm/spout.py
@@ -60,8 +60,10 @@ class SpoutManager:
             raise Exception("spout commit context manager already opened")
         spout_commit = SpoutCommit(self._pipe, marker_filename=self.marker_filename)
         self._has_open_commit = True
-        yield spout_commit
-        spout_commit.close()
+        try:
+            yield spout_commit
+        finally:
+            spout_commit.close()
         self._has_open_commit = False
 
 
--- a/README.md
+++ b/README.md
@@ -45,4 +45,4 @@ However, if for whatever reason you need to know which version of pachyderm core
 
 ## Contributing
 
-This driver is co-maintained by Pachyderm and the community. If you're looking to contribute to the project, this is a fantastic place to get involved. Take a look at [the contributing guide](./contributing.md) for more info.
+This driver is co-maintained by Pachyderm and the community. If you're looking to contribute to the project, this is a fantastic place to get involved. Take a look at [the contributing guide](./contributing.md) for more info (including testing instructions).
--- a/contributing.md
+++ b/contributing.md
@@ -86,11 +86,17 @@ locally it's usually more convenient to run specific tests. To do so:
     * `protobuf>=3.11.2`
     * `grpcio>=1.26.0`
     * `certifi>=2019.11.28`
+* Alternatively, you can use `tox` to create a virtual environment for you:
+  ```
+  mkdir venvdir
+  tox --devenv venvdir -e py38 # one possible environment
+  source ./venvdir/bin/activate # activate python environment
+  ```
 * Install python-pachyderm into the virtualenv: `pip install -e .`
 * Start the cluster to run on `localhost:30650` -- if the cluster is not
 exposed on `localhost`, you can use `pachctl port-forward` to proxy
 connections.
-* Run the test: `py.test tests -k <test name>`
+* Run the test: `python3 -m pytest tests -k <test name>`
 
 ### Linting
 
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -4,6 +4,7 @@
 
 import os
 from contextlib import contextmanager
+from time import sleep
 
 import pytest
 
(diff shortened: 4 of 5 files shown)


src/python_pachyderm/spout.py
```diff
@@ -60,8 +60,10 @@ def commit(self):
             raise Exception("spout commit context manager already opened")
         spout_commit = SpoutCommit(self._pipe, marker_filename=self.marker_filename)
         self._has_open_commit = True
-        yield spout_commit
-        spout_commit.close()
+        try:
+            yield spout_commit
+        finally:
+            spout_commit.close()
         self._has_open_commit = False
```
lukemarsden, line 67: Need to set `_has_open_commit = False` in this case too. (i.e. indent this line)

lukemarsden (changes requested):
Thank you for working on this! It should be easy to add a test for this, just copy the whole test around https://github.com/pachyderm/python-pachyderm/blob/master/tests/test_spout.py#L16 and make that line `raise Exception("boom!")` (with approprate `try: except` in the test to avoid failing the test) and then assert that it's able to further call `with manager.commit()` without error.

Feel free to raise a followup issue for adding the test coverage, but -- it might be easier to write the test than to test it manually.

lukemarsden:
Also, I'm curious why there's an existing test which asserts the opposite of the behavior in this PR:

https://github.com/pachyderm/python-pachyderm/blob/master/tests/test_spout.py#L34-L42

msteffen:
> Also, I'm curious why there's an existing test which asserts the opposite of the behavior in this PR:
> 
> https://github.com/pachyderm/python-pachyderm/blob/master/tests/test_spout.py#L34-L42
> 
> I wonder if (as the existing test suggests) we want the commit to stay open in the case where we error out? If so, how's the user meant to close it?

That is a great question. I made the python fix you suggested, and I'll note in the question about the test in the handoff document.

I can try to get in touch with @ysimonson, but if we can't figure it out, then I think it might make sense to just reverse the test? Under the rationale that if we're wrong, at least this way we'll find out why :shrug:

lukemarsden (approved):

lukemarsden:
I'm fine with the change as-is on the basis of the rationale that once the `with` scope has exited, it ought to have released the resource it represented. Hopefully you'll also need to reverse the test to make the suite pass (`assert not manager._has_open_commit` on https://github.com/pachyderm/python-pachyderm/blob/master/tests/test_spout.py#L42), which would be a minimal change to get this merged. A test which also re-uses the manager to check that it works after an error would be great too.

ysimonson:
I can't recall, but my guess is that it felt "wrong" to close the spout commit on exception. It seems that this change would be more in keeping with context managers though:

```
>>> with open("foo.txt", "w") as f:
...     f.write("hi")
...     raise Exception()
...
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
Exception
>>> f.closed
True
```

msteffen:
This repo seems to be in somewhat unfortunate shape at the moment. I don't think it has any kind of CI set up, and when I ran the full test suite locally to make sure everything passed before I merged, the tests I added pass, but `test_pps` seems to permanently break the Pachyderm 1.11 cluster I'm testing this against for some reason that I don't understand:
```
$ kc logs po/pipeline-test-create-pipeline-from-request-vaj7i6-v1-97vbg --previous user
2020-09-17T09:07:00Z ERROR could not read existing PipelineInfo from PFS: commit a8f11115570549c9bdcff0ae2589ae0d not found in repo __spec__
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xc8 pc=0x209271f]

goroutine 1 [running]:
github.com/pachyderm/pachyderm/src/server/worker/driver.NewDriver(0xc0000da8c0, 0xc0000b6240, 0xc000352f00, 0xc000054040, 0x1d, 0xc00005601a, 0x5, 0x28563f6, 0x1, 0xc00005200f, ...)
	/app/src/server/worker/driver/driver.go:248 +0x93f
github.com/pachyderm/pachyderm/src/server/worker.NewWorker(0xc0000b6240, 0xc000352f00, 0xc000054040, 0x1d, 0xc0000da8c0, 0xc0000500ad, 0x3a, 0xc00005200f, 0x7, 0xc00005601a, ...)
	/app/src/server/worker/worker.go:65 +0xf1
main.do(0x2331a00, 0xc00040f5f0, 0x0, 0x0)
	/app/src/server/cmd/worker/main.go:86 +0x1b9
github.com/pachyderm/pachyderm/src/server/pkg/cmdutil.Main(0x294c960, 0x2331a00, 0xc00040f5f0, 0x0, 0x0, 0x0)
	/app/src/server/pkg/cmdutil/env.go:33 +0x83
main.main()
	/app/src/server/cmd/worker/main.go:42 +0x12e

$ pc list pipeline
commit a8f11115570549c9bdcff0ae2589ae0d not found in repo __spec__
```

I don't think it's safe to merge this PR until we understand the error, unfortunately (I think it's a bug in core Pachyderm, but I don't want to distribute a python-pachyderm package that triggers an open bug), so I think this PR will probably languish until we figure out that issue.

ysimonson:
CI is activated on the repo, it's not showing up in the PR checks due to some issue either with the config or with TravisCI itself: https://travis-ci.org/github/pachyderm/python-pachyderm

ysimonson:
If you run tests locally, make sure to go through the local testing instructions: https://github.com/pachyderm/python-pachyderm/blob/master/contributing.md#testing

msteffen:
> CI is activated on the repo, it's not showing up in the PR checks due to some issue either with the config or with TravisCI itself: https://travis-ci.org/github/pachyderm/python-pachyderm

This is all really helpful! Thanks!

I'll figure out why this PR isn't working in CI before I merge anything then. As far as the PPS issue, I think that's actually a pachyderm bug, and I think fixing that will be separate :shrug:

PR #228, opened 2020-09-30
Support authentication with an OIDC ID token

actgardner:
Add a new auth method to support authenticating with an OIDC ID token (see https://github.com/pachyderm/pachyderm/pull/5292).


msteffen (approved):
LGTM

PR #230, opened 2020-12-23
remove git-lfs

echohack:
Signed-off-by: echohack <git@echohack.app>

Fixes a confusing issue where git-lfs was being used for some simple json data.


mindthevirt (approved):
Just ran through the example and it's working fine now. Thank you @echohack

PR #237, opened 2021-03-15
Remove committed swapfile

actgardner:


msteffen (approved):
LGTM.

Dirty vim users... :joy:

PR #244, opened 2021-03-19
CI is green

actgardner:
Remove the examples until we have 1.12.6, and remove trailing whitespace that causes a linter error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,10 +25,6 @@ jobs:
       env:
         - TOXENV=lint
         - PACHYDERM_VERSION=1.12.4
-    - python: 3.8
-      env:
-        - TOXENV=examples
-        - PACHYDERM_VERSION=1.12.4
 
 install:
 - make ci-install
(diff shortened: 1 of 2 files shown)


msteffen (approved):
LGTM. Two questions:
- Just trying to improve my understanding, why is 1.12.6 a prerequisite for getting the examples to pass? I saw the PR into `pachyderm/pachyderm`—it looked like one change was for the tests in that repo and one was for build pipelines. Is that it?
- Would it make sense to just comment out the examples block, if we're planning to bring it back? (i.e. in case someone else picks this up? Maybe also add a note like "this should work after 1.12.6")

Thanks for taking this on!!

actgardner:
> Just trying to improve my understanding, why is 1.12.6 a prerequisite for getting the examples to pass? I saw the PR into pachyderm/pachyderm—it looked like one change was for the tests in that repo and one was for build pipelines. Is that it?

Yeah, the problem is that opencv-python doesn't work in the Python build pipeline image from Pachyderm 1.12.2 onward. I fixed the image in core pach but we need to wait for a release. 1.12.6 the build image will include libgl so we can re-enable the opencv example.

PR #249, opened 2021-03-22
Rebuild protos for Pachyderm 1.12.5

msteffen:
This looks huge but it's basically a one-line change plus a make command:
- version.json: `1.11.2` -> `1.12.5`
- `make src/python_pachyderm/proto`


actgardner:
I think the script that regenerates the protos is failing to move them from `src/python_pachyderm/protos/client` to just `src/python_pachyderm/protos`? Looking at the current codebase, `src/python_pachyderm/protos/client` doesn't seem to exist? I'm not 100% on how this works though.

msteffen:
> I think the script that regenerates the protos is failing to move them from `src/python_pachyderm/protos/client` to just `src/python_pachyderm/protos`? Looking at the current codebase, `src/python_pachyderm/protos/client` doesn't seem to exist? I'm not 100% on how this works though.

That makes sense. I started reading through the Makefile and this does seem weird. Doing another pass...

actgardner (approved):

actgardner:
Can you rebuild the docs in this PR just so they include the new generated stubs?

msteffen:
> Can you rebuild the docs in this PR just so they include the new generated stubs?

I may be doing super the wrong thing here, but after spending some time trying to understand it, I couldn't figure out what I was supposed to be doing instead. Basically, from the repo root, I ran `rm -rf ./docs; make docs`. It deleted `docs/python_pachyderm.html` and created a dir `docs/python_pachyderm`, which seems wrong but I wasn't sure what to change. In the end I just did `git add ./docs`.

Do you have any familiarity with pdoc? Should I be doing this differently?

actgardner:
Were you using the pypi pdoc, or yusuf's fork (per the contributing.md)?

I'm OK if you want to revert the doc changes and merge as-is, I have pdoc set up on my machine.

msteffen:
> Were you using the pypi pdoc, or yusuf's fork (per the contributing.md)?
> 
> I'm OK if you want to revert the doc changes and merge as-is, I have pdoc set up on my machine.

User error :grimacing: . I didn't see the note about 'make docs' in contributing.md. I installed the custom pdoc and re-ran, and now it looks right (just have the one html file)

actgardner:
Awesome, looks good 👍

.gitignore
```diff
@@ -10,6 +10,9 @@ __pycache__/
 *~
 .DS_Store
 
+# ignore vim temporary files
+.*.sw*
```
actgardner, line 14: 🙏

src/python_pachyderm/mixin/auth.py
```diff
@@ -18,14 +18,24 @@ def activate_auth(self, subject, github_token=None):
...
+        * `root_token`: An optional string. If specified, this string becomes
+          the Pachyderm cluster root user's token. Currently this is only used
+          for testing and Pachyderm internals (migration), so we're avoiding
+          support for this field in the python-pachyderm client until we find a
+          use for them (feel free to file an issue in
+          github.com/pachyderm/pachyderm)
+        """
+        # TODO(msteffen): Add support for branch triggers
+        if trigger is not None:
+            raise NotImplementedError('branch triggers are not supported yet.')
```
actgardner, line 36: This seems like a merge artifact?

msteffen: This was actually a copy and paste error. The intent (which I actually meant to ask for your feedback on) was not to support the `root_token` argument in python_pachyderm. My intuition was that since that was mostly for extract/restore and testing (similar to `extract/restore_auth_token()`), it would be safer to leave it out.

I fixed this error block, though if you think it makes the most sense to just remove it and support `root_token` in python-pachyderm, I'm happy to do that as well.

msteffen: (on doing another pass, I realized there were a handful of careless errors in this PR—not passing arguments through correctly and things like that. I proofread it, though, and _think_ it should be in good shape now)

Issue #255, opened 2021-05-03
Make `repo@commit:path` syntax work in the PFS mixin.


msteffen:
Pachyderm adopted this syntax several releases ago, but python-pachyderm has not yet been updated to support it.

albscui:
Do we want to change the existing behaviour from something like

```
source_file = client.get_file("montage/master", "/montage.png")
```

To

```
source_file = client.get_file("montage@master:/montage.png")
```

Or do we want to support both ways?

jdoliner:
I like standardizing on `<repo>@<commit-or-branch>:<file>` to match pachctl.

albscui:
After some discussion, @JimmyWhitaker and I decided that explicit arguments for `repo`, `commit/branch`, and `path` is a bit nicer from a Python programming perspective.

So instead of `get_file("repo@commit:path")`, our new API would be `get_file("repo", "commit", "path")`.

msteffen:
@albscui That makes sense. If we do this, I think we should drop support for the `repo/commit` syntax: https://github.com/pachyderm/python-pachyderm/blob/9d60ea887a4eeaba45f68a0375ba2640ba48ed01/src/python_pachyderm/mixin/util.py#L7-L11.

For consistency, we might also want to change inspect_file, list_file, walk_file, glob_file, etc.

It might also be worth confirming with LP—I think they use python-pachyderm a bit differently than @JimmyWhitaker. At least, I think Jimmy has said he prefers not to use python-pachyderm interactively in notebooks (correct me if I'm wrong :slightly_smiling_face: ), but IIUC that's the main way LP uses it

albscui:
closing this in favour of https://github.com/pachyderm/python-pachyderm/issues/287

PR #258, opened 2021-05-04
[WIP] Update get_file and make repo arg explicit

albscui:
Currently, `get_file` takes in two arguments to specify a file in PFS,
namely `commit` and `path`. However the `commit` argument is actually a
combo of repo/commit. We want to make the repo argument explicit, so
that there's no confusion between what is a commit and what is a repo.

Relates to #255


jdoliner:
This looks like a good first pass. One thing to think about, you have different calls in the api that take repos i.e just `<repo>` commits / branches i.e `<repo>@<commit-or-branch> and then full files like you have here. So you'll probably want to reuse the various functions. In a few places you also have commands that can accept a commit or a repo. You may want to take a look at the similar code in pachctl to get an idea of how it's structured there since it's somewhat battle tested.

JimmyWhitaker:
Out of curiosity, why not have it have the methods take multiple args rather than a single string that needs to be parsed into an object? 

E.g. 
```python
def get_file(repo: str, branch: str, file: Path): 
    ...
```

I can image some difficulty constructing these strings unless we have a class surrounding it.

albscui:
> Out of curiosity, why not have it have the methods take multiple args rather than a single string that needs to be parsed into an object?
> 
> E.g.
> 
> ```python
> def get_file(repo: str, branch: str, file: Path): 
>     ...
> ```
> 
> I can image some difficulty constructing the pfs reference format (unless we have a class surrounding it that does this for us).

Thanks for the feedback!

I think the scope for this current task is to make the Python client's behaviour consistent with `pachctl`, which uses the `<repo>@<commit-or-branch>:<file>` format.

However, you bring up a good point. What makes sense for a CLI interface might not be optimal for Python's programming interface. This extends across the entire Python Pachyderm project as a whole. IMO, we should be optimizing for "what is most intuitive" for our users. Maybe we can get some user feedback?

I'm drafting up a doc to capture this nuance, and we can discuss the solutions there. https://docs.google.com/document/d/1SlFeWwrKSmjgtBnARDtfYC3wgzdRkmqMbRAD8AqutQU/edit?usp=sharing

PR #259, opened 2021-05-05
Add Black

albscui:
- Ran Black on entire repo
- git blame ignore revision so that this massive change doesn't ruin the git blames
- add pre-commit hook to use Black
- add Black to tox lint


msteffen:
I'm not familiar with Black per se, but code formatting in Go has been great!

One question: I think some of these files (e.g. `src/python_pachyderm/proto/admin/admin_pb2_grpc.py`) are auto-generated by the protobuf compiler. Does it make sense to try and skip those? Or just generate and then format them?

albscui:
> I'm not familiar with Black per se, but code formatting in Go has been great!
> 
> One question: I think some of these files (e.g. `src/python_pachyderm/proto/admin/admin_pb2_grpc.py`) are auto-generated by the protobuf compiler. Does it make sense to try and skip those? Or just generate and then format them?

Good question! I was debating to skip or not to skip. In the end, I decided to format them, because I don't want to introduce unnecessary exceptions to the rule.

The new workflow is like:

- write/update existing code, or generate gRPC code
- on each commit, a pre-commit hook runs Black so we don't have to manually run it every time
- in CI, tox runs Black to check if code conforms to Black's style

With respect to `Black`, I think it adopts the Go formatter's philosophy of being opinionated and not having a lot of options, but works really well out of the box.

jrockway:
This is awesome!

It is very reasonable to not autoformat or lint generated code, specifically the protobufs.  I think that when I regenerate protos, I just want the diff to be from changes the proto compiler is making -- that way I can pretty quickly tell that I used the wrong version, or it was updated dramatically, or something unexpected like that.  If you mix formatting changes in with that, it can be too much.  (My first debugging step when someone says they're having proto trouble is to regenerate them and check the diffs.  If there are no diffs, I know that they did the same thing I did, which is a good state to be in when debugging.  If there is mechanical formatting applied, now there are two variables -- the proto compiler, and the formatter -- so that means there is twice as many potential problems to troubleshoot.)

We don't gofmt or enforce the style of generated code in the Go projects, so I don't think we should introduce that convention here.  Ultimately, the formatter is there so that two humans writing the same code will produce the same sequence of bytes.  The proto compiler will do that anyway, so there's no need to add an extra layer of indirection.  (Regarding linting, there is no easy way to fix lint errors that the code generator emits, so it's probably not useful to surface them.  Just an annoyance that will annoy people.)

For those reasons, I think that `src/python_pachyderm/protos` should be excluded from autoformatting and lint.

msteffen (approved):
Overall LGTM! I pulled the branch, added a test file, and tried out the precommit process, and it all seems to be pretty seamless!

I'm learning about a lot of tools this PR (e.g. I knew about git pre-commits but have like never used them; I didn't know virtualenv had special support for them), so this is all very cool to me.

Re: generated code, after thinking about it a little:
- To @jrockway's point, if we had to fix lint errors in the generated code manually, it sounds tiring to fix the same generated lint errors after every protobuf change (IIUC, flake8 doesn't format, it just lints, so if there's anything flake8 complains about on top of Black, I do think we should continue to exclude them from flake8. But I don't know if there is).
- On the other hand, my 30s experiment with Black suggested that there's nothing manual, so formatting them just adds one extra command. TBH, I haven't had to diff generated protobuf code too many times, so I don't have a sense for how much Black would change that process.
- My sense is that if it does get annoying, this decision is relatively easy to un-make (just excluse the generated files, rebuild the protos, and commit)

Based on all that, I don't feel compelled the hold the PR up on it. You could try adding a field to one of the protos and doing a diff with and without Black to see how it affects readability if you wanted (which could also be good exercise in rebuilding the protos, which took me a while to figure out)

jrockway (approved):
Agree with @msteffen there.  If we need to change the rules in the future, we can change them.

albscui:
@jrockway I updated the config for Black to exclude the autogenerated protos!

PR #260, opened 2021-05-10
'v7.x alpha' rebased onto 'master minus black'

msteffen:


Makefile
```diff
@@ -23,12 +23,13 @@ src/python_pachyderm/proto: docker-build-proto
...
 		git fetch --all && \
 		git checkout v$(PACHYDERM_VERSION)
-	find ./proto/pachyderm/src/client -regex ".*\.proto" \
+	find ./proto/pachyderm/ -regex ".*\.proto" \
 	| grep -v 'internal' \
 	| xargs tar cf - \
 	| docker run -i pachyderm_python_proto \
 	| tar xf -
-	mv src/python_pachyderm/client src/python_pachyderm/proto
+	rm -rf src/python_pachyderm/proto
```
albscui, line 31: Redundant line, since we already have one above
https://github.com/pachyderm/python-pachyderm/pull/260/files?file-filters%5B%5D=No+extension#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L22

Makefile
```diff
@@ -43,7 +44,7 @@ ci-install:
 	sudo dpkg -i /tmp/pachctl.deb
 	pip install tox tox-travis
 
-ci-setup:
+ci-setup:
```
albscui, line 47:
```suggestion
ci-setup:
```

msteffen (commented):
Still reviewing some of the mixins, but so far LGTM. All comments are nits that I think could be fixed in followup PRs

etc/proto_lint/proto_lint.py
```diff
@@ -66,23 +68,34 @@
...
+        "create_fileset",
     ],
     Service.PPS: [
         # the following are ignored because they're for internal use only
         "activate_auth",
         "create_job",
         "update_job_state",
     ],
+    Service.AUTH: [
+        "get_auth_token",
```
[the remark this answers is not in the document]
albscui: I believe this is part of the `BLACKLISTED_METHODS` list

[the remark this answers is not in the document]
actgardner: `get_auth_token` and `extend_auth_token` are both gone from the RPCs and they should be removed from the proto_lint in the interest of tidiness.

src/python_pachyderm/mixin/auth.py
```diff
@@ -57,64 +39,30 @@ def set_auth_configuration(self, configuration):
...
-    def get_cluster_role_bindings(self):
+    def get_role_binding(self, resource):
         """
         Returns the current set of cluster role bindings.
         """
-        return self._req(Service.AUTH, "GetClusterRoleBindings")
+        return self._req(Service.AUTH, "GetRoleBinding", resource=resource)
 
-    def modify_cluster_role_binding(self, principal, roles=None):
+    def modify_role_binding(self, resource, principal, roles=None):
```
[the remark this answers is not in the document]
actgardner: `resource` can be a reference to the cluster or a specific repo (or in the future a project, group, robot account, etc.). This replaces `get_x_role_binding` where x is the type of resource.

src/python_pachyderm/mixin/pfs.py
```diff
@@ -75,6 +55,26 @@ def read(self, size=-1):
...
+
+    ```
+    source_file = client.get_file("montage/master", "/montage.png")
+    with open("montage.png", "wb") as dest_file:
+        shutil.copyfileobj(source_file, dest_file)
+    ```
+    """
+
+    def __init__(self, stream):
+        f = tarfile.open(fileobj=stream, mode='r|*')
```
albscui, line 71: Maybe provide a brief explanation to why we are using `tarfile`?

Makefile
msteffen: Removing the extra `rm -rf` above meant GitHub couldn't apply this :-(. Added it in a followup commit

src/python_pachyderm/mixin/pfs.py
albscui: This is because GetFile can return an arbitrary directory tree. Tar is the format we decided on for serializing this structure.

src/python_pachyderm/mixin/pfs.py
```diff
@@ -75,6 +55,29 @@ def read(self, size=-1):
         return b"".join(buf)
 
 
+class PFSFile:
```
albscui, line 58: We should also add a `close()` method.

etc/proto_lint/proto_lint.py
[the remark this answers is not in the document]
msteffen: @albscui Oh whoops, I see it now. I think your point is that it doesn't make a difference, which makes sense. I did just remove it—hopefully that's fine.

tests/test_auth.py
```diff
@@ -10,114 +10,83 @@
 
 import python_pachyderm
 from tests import util
-
+from python_pachyderm.proto.auth import auth_pb2
+from python_pachyderm.proto.identity import identity_pb2
 
 @contextmanager
```
albscui, line 16: Not a blocker, but instead of `@contextmanager`, we could try using `@pytest.fixture`, which provides some nice features like safe teardowns.

tests/test_pfs.py
```diff
@@ -39,15 +39,15 @@ def test_delete_non_existent_repo():
     assert len(client.list_repo()) == orig_repo_count
 
 
-def test_delete_all_repos():
-    client = python_pachyderm.Client()
+#def test_delete_all_repos():
```
albscui, line 42: We should just remove these if they are not being used for testing.

tests/test_pfs.py
```diff
@@ -357,25 +327,24 @@ def test_inspect_commit():
...
-
-    with client.commit(repo_name, 'master') as c:
-        pass
-
-    commits = list(client.list_commit(repo_name))
-    assert len(commits) == 1
-    client.delete_commit("{}/master".format(repo_name))
-    commits = list(client.list_commit(repo_name))
-    assert len(commits) == 0
+#def test_delete_commit():
```
albscui, line 337: We should remove these if they are not being used.

tests/test_pfs.py
```diff
@@ -521,10 +487,10 @@ def test_diff_file():
...
-    assert diff.new_files[0].file.path == "file1.dat"
-    assert diff.old_files[0].file.path == "file2.dat"
+    diff = list(client.diff_file(new_commit, "file1.dat", old_commit, "file2.dat"))
+    assert diff[0].new_file.file.path == "/file1.dat"
+    assert diff[1].old_file.file.path == "/file2.dat"
 
-    diff = client.diff_file(new_commit, "file1.dat")
-    assert diff.new_files[0].file.path == "file1.dat"
-    assert diff.old_files[0].file.path == "file1.dat"
+    #diff = list(client.diff_file(new_commit, "file1.dat"))
```
albscui, line 494: Remove these comments as well.

albscui (commented):
Just a few cleanups for commented out code. The pytest fixtures is not a blocker for this PR, just noting it here to remind myself to maybe add it later.

Overall, LGTM.

tests/test_pps.py
```diff
@@ -78,23 +78,18 @@ def test_datums():
...
+    datum = sandbox.client.inspect_datum(job_id, datums[0].datum.id)
     assert datum.state == python_pachyderm.DatumState.SUCCESS.value
 
-    # Skip this check in >=1.11.0, due to a bug:
-    # https://github.com/pachyderm/pachyderm/issues/5123
-    # TODO: remove this check once the bug is fixed
-    if util.test_pachyderm_version() < (1, 11, 0):
-        # Just ensure this doesn't raise an exception
-        sandbox.client.restart_datum(job_id)
+    # TODO: This requires catching a datum while it's being processed
```
msteffen, line 84: @Tryneus Hey Marc, I think I'm missing some context, but just from https://github.com/pachyderm/pachyderm/issues/5123, what would normally go in the `DataFilters` field of a cancel request? Why does the datum have to be in progress?

tests/test_pfs.py
msteffen: @actgardner Thanks for answering all of these questions—it looks like this function is still in the PFS mixin. Does the test not work in 2.0 or is it just flaky or something?

If the issue is just that it's flaky, we should add a todo and fix it. Otherwise it seems like we should remove the corresponding method from the client

tests/test_pfs.py
```diff
@@ -148,45 +148,10 @@ def test_put_file_bytes_bytestring():
     commit_infos = list(client.list_commit(repo_name))
     assert len(commit_infos) == 1
     assert commit_infos[0].commit.id == c.id
-    files = list(client.list_file('{}/{}'.format(repo_name, c.id), '.'))
+    files = list(client.list_file('{}/{}'.format(repo_name, c.id), ''))
```
msteffen, line 151: Not really relevant to this line, but we should change python-pachyderm to use the same short format as pachctl: `repo@commit[:file]` rather than `repo/commit`

msteffen (commented):
A few more hopefully quick questions...

tests/test_pps.py
Tryneus: It looks like there's several things you can specify as a filter: https://github.com/pachyderm/pachyderm/blob/a2d9ffbff33597be6038decc7c022bdc953dc8b6/src/server/worker/common/common.go#L53

A matching datum does not have to be in-progress, we will just return a `Success: false` if no action was taken.  I don't believe there is any way to test this without a race unless you do something super fancy with a specialized workload.  In this case, I would expect `Success` to always be false because this test waits for the job to finish before attempting to restart it.  `RestartDatum` should never error unless there is a transport issue, from what I can tell.

tests/test_pfs.py
msteffen: Nevermind, we already have https://github.com/pachyderm/python-pachyderm/issues/255. Marking resolved.

msteffen (commented):
Few more TODOs

src/python_pachyderm/mixin/pfs.py
```diff
@@ -75,6 +55,29 @@ def read(self, size=-1):
...
+        shutil.copyfileobj(source_file, dest_file)
+    ```
+    """
+
+    def __init__(self, stream):
+        # Pachyderm's GetFile API returns its result (which may include several
+        # files, e.g. when getting a directory) as a tar stream--untar the
+        # response byte stream as we receive it from GetFile.
+        f = tarfile.open(fileobj=stream, mode='r|*')
+        self._file = f.extractfile(f.next())
```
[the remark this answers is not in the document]
msteffen: I worry about the case where `GetFile` is called with multiple directories, but it seems like that's high-level incompatible with making `pandas` and `csv` work nicely, which I gather this aims to do. I don't know what the best way to deal with the split is, though my intuition is that all files should be concatenated into one super-huge dataframe (there is precident for this in Pachyderm 1.x—we should discuss later).

In any case, we can investigate this further as a next step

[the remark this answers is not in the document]
albscui: Yea there's a bit of an impedance mismatch here. I would say the average use case for `GetFile` is exactly to get a single file.

albscui (approved):
Just rubber stamping this to unblock future development work.

PR #262, opened 2021-05-12
Pull out protobuf build changes

msteffen:


albscui:
I think this is missing one of the generated protobufs that should be found at `src/python_pachyderm/proto/admin/v1_12`

There are also some more `etc/...` files that should be deleted as part of these commits:
- https://github.com/pachyderm/python-pachyderm/pull/260/commits/43c27a3fb690eafefc9706231039d663de6de18d
- https://github.com/pachyderm/python-pachyderm/pull/260/commits/ebff21d5cf27e41f2f51c51bee92baa8cc823bc4

albscui (approved):
I pushed some changes to get CI to pass, but overall, LGTM!

PR #264, opened 2021-05-13
Upgrade kubernetes version in CI

msteffen:


albscui (approved):

albscui:
We should also merge this into master once we are ready.

PR #270, opened 2021-05-19
[skip ci] Add proto_lint to pre-commit

albscui:


msteffen (commented):
Nice!

Would it be possible to run this in CI in addition to running it as in the commit hook (IIUC, we may not be able to guarantee that contributors actually run the commit hooks)?

albscui:
> Nice!
> 
> Would it be possible to run this in CI in addition to running it as in the commit hook (IIUC, we may not be able to guarantee that contributors actually run the commit hooks)?

Hmm, I'm pretty sure we already run proto_lint in CI right now.

msteffen:
> Hmm, I'm pretty sure we already run proto_lint in CI right now.

Whoops, my bad! I see it now :joy:

msteffen (approved):

PR #278, opened 2021-05-20
Improve documentation and docs generation

albscui:
- better example in README.md
- updated how we generate API docs with pdoc


albscui:
To-Do
- implement a script that runs in a Docker container which removes the generated protos from src so that they don't pollute the generated documentation

msteffen (approved):
LGTM

PR #281, opened 2021-05-27
Improve proto generation and cleanup old protos

albscui:
While attempting to move proto generation to pachyderm/pachyderm, we've learned a few things and improved the scripts for this process. However, we decided that proto generation and packaging is a bigger project that involves the organization as a whole, so we won't be migrating proto generation from python-pachyderm to pachyderm/pachyderm atm. Related issue here https://github.com/pachyderm/pachyderm/issues/6116.

However, we can still benefit from this exercise and apply some of the improvements:

- Simplified and more valid `proto/run` for generating protos
- Parameterize the proto generator's output dir
  - this will allow us to generate different protos based on different versions of pachyderm/pachyderm
- Slimmer Docker image using `python:3.6-slim`

We are also adding a new `python_pachyderm.proto.v2` namespace to store the protos generated from pachyderm v2.x

We **removed** existing generated protos from `python_pachyderm/proto`.

### Next steps

- We can generate a set of protos based off of pachyderm/pachyderm v1.x
- Update `python_pachyderm/__init__.py` to defer importing a specific proto version, so that `python_pachyderm.proto.v1` and `python_pachyderm.proto.v2` can co-exist in the same repo

Relates to https://github.com/pachyderm/python-pachyderm/issues/282


msteffen (approved):
LGTM!

PR #286, opened 2021-06-01
Parity with pachyderm v2.0.0-alpha.20

albscui:


.travis.yml
```diff
@@ -15,19 +15,18 @@ env:
   global:
     - PATH=$HOME/cached-deps:$PATH
   jobs:
-    - PACHYDERM_VERSION=2.0.0-alpha.12
-
+    - PACHYDERM_VERSION=$(jq -r .pachyderm version.json)
```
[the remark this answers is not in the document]
msteffen: It's a bit weird that we set `PACHYDERM_VERSION` to the same value everywhere—it seems like we should just take it out, but maybe it makes sense to leave it for when 2.1 comes out, or if we add back 1.x support?

src/python_pachyderm/client.py
```diff
@@ -61,6 +61,7 @@ def __init__(
         root_certs=None,
         transaction_id=None,
         tls=None,
+        max_message_length=(1024 ** 2) * 20,
```
msteffen, line 64: I think it might make sense to make this a constant, since it's a constant in the server. Or do you think people might need to change it?

src/python_pachyderm/mixin/auth.py
```diff
@@ -165,7 +165,7 @@ def modify_members(self, group, add=None, remove=None):
             remove=remove or [],
         )
 
-    def get_groups(self, username=None):
+    def get_groups(self):
         """
         Gets which groups the given `username` belongs to. Returns a list of
```
msteffen, line 170: I think the docstring might need to be updated

src/python_pachyderm/mixin/auth.py
```diff
@@ -174,7 +174,7 @@ def get_groups(self, username=None):
 
         `username`: A string.
```
msteffen, line 175: I think the docstring might need to be updated.

src/python_pachyderm/mixin/util.py
```diff
@@ -1,18 +1,35 @@
...
+
+    repo, branch, commit = None, None, None
+    if isinstance(src, (tuple, list)) and len(src) == 2:
+        repo, branch_or_commit = src
     elif isinstance(src, str):
-        repo_name, commit_id = src.split("/", 1)
-        return pfs_proto.Commit(repo=pfs_proto.Repo(name=repo_name), id=commit_id)
+        repo, branch_or_commit = src.split("/", 1)
+    if uuid_re.match(branch_or_commit) or not valid_branch_re.match(branch_or_commit):
+        commit = branch_or_commit
```
msteffen, line 23: It seems like, if the code takes this branch, then `branch` is `None`, which I think is invalid (e.g. for `ListFile`—I think if `pfs_proto.Commit.id` is `None`, then the most recent commit is assumed, but if `pfs_proto.Commit.branch.name` is `None`, then that's an error)

src/python_pachyderm/mixin/util.py
```diff
@@ -1,18 +1,35 @@
...
+        branch = branch_or_commit
+
+    if repo is None:
+        raise ValueError("Repo cannot be empty")
 
-    if not allow_just_repo:
-        raise ValueError("Invalid commit type")
-    return pfs_proto.Commit(repo=pfs_proto.Repo(name=src))
+    return pfs_proto.Commit(
+        id=commit,
```
msteffen, line 31: Does this parse `src` differently than the old version, or is it just the output that changed?

I think it might make sense rid of this and focus on moving to the `(repo, branch, commit)` syntax in the calls, since I think there's no way to use this to see commits other than the most recent

src/python_pachyderm/util.py
```diff
@@ -69,7 +70,7 @@ def put_files(client, source_path, commit, dest_path, **kwargs):
 
 
 def create_python_pipeline(
-    client,
+    client: Client,
```
msteffen, line 73: Just curious, since I don't understand python type annotations that well: does this add functionality or is it just a convenience for users?

tests/test_debug.py
```diff
@@ -2,10 +2,7 @@
 
 """Tests debug-related functionality"""
 
-import pytest
```
msteffen, line 5: Nice!

tests/test_pfs.py
```diff
@@ -39,51 +39,45 @@ def test_delete_non_existent_repo():
...
+    assert len(client.list_repo()) == 0
 
 
 def test_start_commit():
     client, repo_name = sandbox("start_commit")
 
     commit = client.start_commit(repo_name, "master")
-    assert commit.repo.name == repo_name
-
-    commit = client.start_commit(repo_name, None)
```
msteffen, line 62: Just curious: it's a server-side change that `master` is no longer assumed?

msteffen (approved):
Thanks for keeping up with all these changes!

I made a handful of comments, but if you need to get this in to keep `python-pachyderm` up to date and fix the issues in followups, that makes sense to me.

tests/test_pfs.py
albscui: Yea, the server now requires branch name.

https://github.com/pachyderm/pachyderm/blob/65eaab4358a8af6b11aa783ab66c6f8c7c0d4db9/src/server/pfs/server/driver.go#L532

src/python_pachyderm/util.py
albscui: Python type hints don't actually enforce types in the same way that statically typed languages do. But this helps IDEs and linters. In this example, the IDE now understands that client is of type `Client`, and can highlight its members, as well provide auto suggestions.

Type hints also help with documenting code. Currently, we specify the arg type in our docstring, but this is adhoc, and doc generators are not necessarily going to pick this up automatically. Python type hints standardizes this, so doc generators can parse them more easily.

If we decide to add type hints throughout our project, it would be quite a massive undertaking though. I think moving forward, we can try to add type hints incrementally, and definitely for new functions.

src/python_pachyderm/mixin/util.py
albscui: Yes this was a bit confusing to me at first. However, I think the server can take in an empty branch name. At least the way we currently parse inputs to `pachctl` suggests this.

https://github.com/pachyderm/pachyderm/blob/3b69deea92a21ba5877b8792b99ba5acfd1258e0/src/internal/cmdutil/cobra.go#L156-L159

src/python_pachyderm/mixin/util.py
albscui: Only the output has changed. I can also drop support for `"repo/commit-or-branch"` syntax here.

.travis.yml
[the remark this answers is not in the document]
msteffen: For posterity: we used to set this to multiple values in order to run the python-pachyderm test suite against multiple versions of pachyderm. We may need to do that again (especially if we decide to try and support both Pachyderm 1.x and pachyderm 2.x in the same python-pachyderm release) but for now, we're only running the test suite against the latest 2.0 Pachyderm release

src/python_pachyderm/mixin/util.py
msteffen: Interesting. I may ping Marc to ask what the semantics of this is—I'm not sure what pachyderm does with a commit but no branch (especially since the test suite needed to start setting `"master"` everywhere)

PR #290, opened 2021-06-12
Fix broken tests by explicitly importing protos

albscui:
Master is broken because we no longer dynamically import protobuf objects.
https://github.com/pachyderm/python-pachyderm/blob/d8c4035033965a07204995d642d7225b7d80ef63/src/python_pachyderm/__init__.py#L95-L103

This was intentional because we don't want doc parsers to include these, but also we don't want to leak protobuf details in general.


msteffen (approved):
LGTM

PR #292, opened 2021-06-16
Make `Commit` objects more Pythonic

### LINKED ISSUE #287: [PFS mixin] Update interface for working with commits
# Background

Global IDs: https://github.com/pachyderm/rfcs/blob/global_ids/global_ids.md

The TLDR:
- A user can specify a commit in 3 ways
  - `repo, branch`
  - `repo, commit_id`
  - `repo, branch, commit_id`

# Problem:

Although Pachyderm uses gRPC + protobuf, the client should hide these details. Currently, our API leaks protobuf details by taking in protobuf objects and returning them.

# Solution
The solution is to implement adapters that are more Pythonic, but can be translated from and to protobuf objects.

# Design

Implement a `Commit` type that acts as the Pythonic version of `pfs_pb2.Commit`

```python
class Commit(NamedTuple):
    repo: str
    branch: str = None
    id: str = None
    repo_type: str = "user"

    def _to_pb(self) -> pfs_pb2.Commit:
    
    @staticmethod
    def _from_pb(commit: pfs_pb2.Commit) -> "Commit":

```

However, our APIs don't need a `Commit` object explicitly. In most cases, a `tuple` or `dict` can fully capture the information required for a `Commit`.

```python
# A CommitType can be a Commit, tuple, or dict
CommitType = Union[Commit, tuple, dict]

# Example interfaces
start_commit(repo, branch, parent: CommitType) -> Commit
flush_commit(commits: list[CommitType])
inspect_commit(commit: CommitType)
get_file(commit: CommitType, path: str)
```

albscui:
Resolves #287


src/python_pachyderm/pfs.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Helper function to convert objects that represent a Commit query into a
+    protobuf Commit object.
+    """
+    if isinstance(commit, pfs_pb2.Commit):
+        return commit
+    if isinstance(commit, Commit):
+        return commit.to_pb()
+    if isinstance(commit, tuple):
+        repo, branch, commit_id = None, None, None
```
smalyala: I see that the assumption that `len(tuple)` is either `2` or `3` matches the code from the now-deleted `util.py` file. However, why can't `user` to be specified if `commit` is a `tuple` type but can if commit is a `dict` type?

tests/test_pfs.py
```diff
@@ -98,7 +98,16 @@ def test_start_commit_fork():
     assert "patch" in branches
 
 
-@pytest.mark.parametrize("commit_arg", ["commit_obj", "(repo, commit_id)"])
+@pytest.mark.parametrize(
```
smalyala, line 101: Is it worth adding tests for tuples and dictionaries with field `type`?

src/python_pachyderm/pfs.py
smalyala: Ignore this, this is addressed in the next commit.

smalyala (approved):
Just one comment on tests

Issue #293, opened 2021-06-16
Support putting individual files


msteffen:
Currently, the python-pachyderm helper function for putting files, `put_files`, doesn't work if `source_path` is an individual file. This is because the implementation calls `os.walk`, which returns an empty list if its argument is a regular file (rather than a directory)

Since putting an individual file is by far the more common use of e.g. `pachctl put file`, we'd like to support that case in python as well. Thus, we'd like to make two changes:

1) call `os.stat` on the `source_path` passed to `put_files`, and if that argument is a regular file, just upload it directly
2) Add a function `put_file` that is similar to `put_files`, but requires its argument to be a regular file.

smalyala:
Decided to consolidate new `put_file` functionality into already existing `put_files` because logic is very similar. Creating a new function seems unnecessary from a user perspective.

PR #294, opened 2021-06-22
[PFS] Add functionality to put_files to support putting individual files

smalyala:
#293 
- `put_files` checks if source_path is individual file; uploads if it is file
- `put_files` checks if source_path is directory; uploads files recursively if it is directory
- otherwise throws an error
- added test case


albscui (approved):
LGTM!

PR #299, opened 2021-06-28
Refactor new_from_config logic into Client default constructor

### LINKED ISSUE #274: `Client()` should use `new_from_config()` (and check `/pachyderm/config.json`) by default
Currently, when using python_pachyderm in e.g. a service spout, users have to manually track down the pach config that is generated and placed in the service spout container and then manually initialize a python_pachyderm client with `NewFromConfig`. Instead, `python_pachyderm` should search for configs in the same places and in the same order as `pachctl`, so that initializing a python_pachyderm client is never any harder than connecting with `pachctl`

(One note: `pachctl` has port-forwarding built in, and automatically uses kubernetes port-forwarding when no config is available. It's not clear to me if/how to do this in python-pachyderm, but this is one place where it may not be feasible to achieve parity. This is the only exception I can think of, though)

smalyala:
Closes #274 

Decided not to raise error if only one of `host` or `port` is set because seems possible where a user might only specify one and expect the default for the other value to be used. Additionally, some of the tests covered this case, validating that a user specifying only one of the two is a situation to support rather than flag.


src/python_pachyderm/client.py
```diff
@@ -62,24 +62,55 @@ def __init__(
...
         * `root_certs`:  The PEM-encoded root certificates as byte string.
         * `transaction_id`: The ID of the transaction to run operations on.
         * `tls`: Specifies whether TLS should be used. If `root_certs` are
         specified, they are used; otherwise, we use the certs provided by
         certifi.
+        * `use_default_host`: Boolean for replicating `pachctl` behavior 
+        of searching for config. Default is True.
         """
 
+        # replicate pachctl behavior to searching for config if host and port are unset
```
albscui, line 89: I think we should try to factor this out into a helper function and keep the __init__ clean.

src/python_pachyderm/client.py
```diff
@@ -172,80 +194,157 @@ def new_from_pachd_address(
...
+    @staticmethod
+    def check_for_config():
+        """
+        Checks for config.
+
+        Returns:
+        * `j`: The config as a json object.
+        """
+
+        j = Client.check_pach_config_env_var()
```
smalyala, line 252: Do you know a slicker way to do these checks one by one?

albscui: I think this is slick enough 😉
You could try something like

```python
configs = [Client.check_pach_config_env_var(), Client.check_pach_config_spout(), Client.check_pach_config_local()]
return next((config for config in configs if config), "boo no configs")
```

src/python_pachyderm/client.py
```diff
@@ -172,80 +194,157 @@ def new_from_pachd_address(
...
         try:
-            context = j["v2"]["contexts"][active_context]
+            context = config["v2"]["contexts"][active_context]
         except KeyError:
             raise ConfigError("missing active context '{}'".format(active_context))
 
+        return context
+
+    @staticmethod
+    def _parse_config(config):
```
albscui, line 326: maybe add a test for this as well?

smalyala: added

albscui (approved):
LGTM!

Issue #300, opened 2021-06-28
Migrate from Travis to CircleCI


albscui:
Core Pachyderm has fully migrated to CircleCI, and so should we!

- [x] Set up CircleCI on master
- [x] Back port CircleCI for v6.x
- [x] Stop using Travis

PR #301, opened 2021-06-29
Run tests on CircleCI

actgardner:
This moves CI for master from Travis to CircleCI (https://github.com/pachyderm/python-pachyderm/issues/300)

The py36 target is failing but it also appears to be failing with the same error in Travis.

Things to consider in the future:
- Managing multiple Python versions. I'm not an expert here, I kind of assumed tox would create the virtualenvs and install whetever version of Python we needed. Right now this uses the deadsnakes PPA to install py36 and py39 on the system path.
- Getting rid of port forwarding. Previously we ran minikube with the none driver, which meant nodeports could be accessed at localhost. This PR uses the Docker driver for consistency with the pach repo, so the nodeports are now at the docker IP (`minikube ip`). For now this PR does a port-forward, but ideally the tests (and maybe the client constructor) would load the pachctl config instead of assuming localhost.


albscui (approved):
LGTM! And thanks so much for setting this up!!

actgardner:
No worries - after it's soaked for a week or so we should make sure to turn off Travis so we're not paying for both.

PR #302, opened 2021-06-29
Backport circleci config to v6.x

actgardner:
This is just cherry-picking the change from #301 to the v6.x branch. It only runs against the version of pachyderm in version.json (1.13.2), but we could adapt it to do multiple pach versions instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,48 @@
+version: 2.1
+
+jobs:
+  tox:
+    resource_class: large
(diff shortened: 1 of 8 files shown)


albscui (approved):
LGTM!

I updated the tox.ini to use py36 and py39. However, it's a bit of a mystery how tox actually works, since CircleCI didn't fail the first time when there was a mismatch between the tox.ini and the actual python interpreters installed.

PR #305, opened 2021-07-19
Parity with 2.0.0-beta.2

smalyala:
- Changing default Python values for grpc methods to match `pachctl` will be in a different PR


src/python_pachyderm/mixin/pps.py
```diff
@@ -301,60 +327,52 @@ def create_pipeline_from_request(self, req):
...
-            * 2: etc.
-            * -1: Return jobs from all historical versions.
+        * `details`: An optional bool that indicates to return details
+        on the pipeline.
         """
-        pipeline = pps_proto.Pipeline(name=pipeline_name)
-
-        if history is None:
-            return self._req(Service.PPS, "InspectPipeline", pipeline=pipeline)
-        else:
```
smalyala, line 323: Moved this code from `else` block into the list_pipeline method.

tests/test_pps.py
```diff
@@ -45,6 +45,18 @@ def test_list_job():
...
+def test_list_job_set():
+    sandbox = Sandbox("list_job_set1")
+    sandbox.wait()
+    client = sandbox.client
+
+    commit, _, _ = util.create_test_pipeline(client, "list_job_set2")
+    client.wait_commit(commit.id)
+
+    jobs = list(client.list_job_set())
+    assert len(jobs) == 4
```
albscui, line 57: Why 4 instead of 2?

albscui (commented):
I think for the PPS tests that are hanging, we should just skip them for now, or deliberately fail them so that they don't hang in CI.

src/python_pachyderm/mixin/pps.py
```diff
@@ -301,60 +327,52 @@ def create_pipeline_from_request(self, req):
...
+            pipeline=pps_proto.Pipeline(name=pipeline_name),
+            details=details,
+        )
 
-    def list_pipeline(self, history=None, allow_incomplete=None, jqFilter=None):
+    def list_pipeline(
+        self, pipeline_name=None, history=None, details=None, jqFilter=None
+    ):
         """
         Lists pipelines. Returns a `PipelineInfos` object.
```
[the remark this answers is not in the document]
smalyala: good catch

tests/test_auth.py
```diff
@@ -82,6 +82,18 @@ def test_who_am_i(client):
     assert client.who_am_i().username == "pach:root"
 
 
+@util.skip_if_no_enterprise()
+def test_get_roles_for_permission(client):
+    # Checks built-in roles
+    roles = client.get_roles_for_permission(auth_proto.Permission.REPO_READ)
+    assert len(roles.roles) == 4
```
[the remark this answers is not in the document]
smalyala: modified so it's not dependent on a hardcoded # of roles

tests/test_pfs.py
```diff
@@ -415,6 +415,25 @@ def test_subscribe_commit():
...
+    python_pachyderm.Client().delete_all_repos()
+
+    client1, repo_name1 = sandbox("list_commit_set_1")
+
+    with client1.commit(repo_name1, "master"):
+        pass
+    with client1.commit(repo_name1, "master"):
+        pass
+
+    client2, repo_name2 = sandbox("list_commit_set_2")
```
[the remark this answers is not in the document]
smalyala: fixed

tests/test_pps.py
smalyala: Seems like in the scenario where an input and output repo are created, followed by a pipeline, there's an initial empty commit that propagates through. The user commit then becomes the second commit. https://pachyderm.slack.com/archives/C01M3RQNVCP/p1626725668243300

etc/testing/circle/deploy-pachyderm.sh
```diff
@@ -3,9 +3,12 @@
...
+helm repo add pachyderm https://pachyderm.github.io/helmchart
+helm repo update
+helm install pachd pachyderm/pachyderm --set deployTarget=LOCAL --version ${PACHYDERM_VERSION}
 
-echo 'y' | pachctl deploy local
 kubectl wait --for=condition=available deployment -l app=pachd --timeout=5m
 pachctl version
-
-pachctl config update context "$(pachctl config get active-context)" --pachd-address="$(minikube ip):30650"
+rm -f ~/.pachyderm/config.json
```
albscui, line 14: Does this file persist in CircleCI across jobs?

albscui (approved):
Awesome, LGTM!

etc/testing/circle/deploy-pachyderm.sh
smalyala: Updating the `pachd-address` here is what was causing those tests to fail so figured I'd get rid of that file in case it does exist. Just pushed another commit getting rid of the `rm -f ~/.pachyderm/config.json` as it's not needed.

PR #308, opened 2021-08-03
Update Spouts for v2

### LINKED ISSUE #276: Does SpoutManager still work, post "new spouts"?
Pachyderm recently changed the UX for spouts (TODO: post issue/PR). This issue is to:
1. Review the new UX, which I only sort of understand—I think pachd injects a copy of `pachctl` into the spout container via the `init` container, as well a pach config connecting it to the spout's sidecar, and then users use that to emit data from the spout.
    - I think the old UX was that users would write a tar stream to `/pfs/out`, but this was unreliable due to wide variation in tar implementations/libraries
2. Make sure python-pachyderm's spout code (particularly the SpoutManager library) still works and still aligns with the UX

smalyala:
- Remove SpoutsManager since no longer using tar and pipe
- Update spouts example, which reproduces the Spouts101 ex in core pach 
- Update spouts tests

spout example files changed as so:
- `examples/spout/producer/main.py` -> `examples/spout/consumer/poll_consumer.py` 
-  `examples/spout/consumer/main.py` -> `examples/spout/processor/log_messages.py`
- `examples/spout/producer/` folder -> `examples/spout/consumer/` folder
- `examples/spout/consumer/` folder -> `examples/spout/processor/` folder

Will add README to v6
Will update Spouts101 example in core pach repo since it uses python-pachyderm

cc @nadegepepin, @msteffen 
closes #276


nadegepepin:
It took me a little time to understand why you would rename `producer` to `consumer` until I realized that it had been badly copied from the original example where the `consumer` was the "Polling Consumer" in the sense of the [Design Pattern](https://www.enterpriseintegrationpatterns.com/patterns/messaging/PollingConsumer.html), receiving the messages from a third party system (say a queue), and the `processor` was processing those received messages. BTW `processor` was a deliberate choice of name. A `producer` does not fit that Pattern. Anyhow,  you put things back to where they were supposed to be. Thank you.

 I have 2 general remarks but overall, LGTM!:
 - It would be great to have 1 single DockerFile and MakeFile (ie: one image). It is a didactic example, therefore should be oversimplified and easy to use. I understand why you split both but I believe that one single Docker image with each of the 2 pipelines calling a transform.cmd that is different, works well. I am open to discussion, this is a personal opinion. 
 - When the original spout101 is adapted to v2, wouldn't it make sense to remove the consumer and processor .py from python-pachyderm to avoid any redundancy? This way, you simply draw attention to the use of the python client to create your pipelines rather than the CLI. Although, it might be your plan.
 
 Anyway, thank you for this adaptation and the added "reducer"!

smalyala:
Sounds good, I will make both of these changes in the next iteration. I will move the example code to the Pachyderm repo as well as consolidate the Dockerfile and Makefile.

nadegepepin:
Also: We discourage the use of the "Latest" TAG in our Docker Images (Because of our Pull Policy: "If Present"). A good practice is to build the image with a TAG and reference it in the pipeline transform section. In Pachyderm's main repo, the tag used is the version of Pachyderm the example was last tested on. ex: 1.12.2

nadegepepin (approved):
See comments.

PR #311, opened 2021-08-05
Migrate spouts example to Pachyderm Spouts101 folder

smalyala:
To avoid redundancy, we moved all the python code to the Spouts101 folder (https://github.com/pachyderm/pachyderm/pull/6693). The pipelines in the python spouts example use the docker image created in the Pachyderm Spouts101 example.


nadegepepin (approved):
Awesome. LGTM!

PR #326, opened 2021-08-20
Revert list_pipeline and inspect_pipeline

### LINKED ISSUE #314: PPS: Question: Should `list_pipeline` include an optional parameter for `pipeline_name`?
Hello,

This morning I was looking at this method due to some customer questions, and it struck me as very weird that this optional parameter was added for `list_pipeline`. What exactly is this parameter doing? It seems like circular logic that you'd need to know the pipeline name before you got a list of all the pipelines (and that `inspect` should give you what you want once you know the pipeline name).

I also found it odd that this was NOT included in 6.x python-pachyderm, and instead was a bug(?) in the docstring. Is it possible that this was accidentally carried forward and doesn't actually match what the actual PPS API should be doing?

See also: #312

smalyala:
Closes #314 

Reverted `list_pipeline()` and `inspect_pipeline()` to what they were before. It's confusing for `list_pipeline()` to take an arg for `pipeline_name` since the point of the function is to list pipelines. The reason why the RPC is designed to take that arg is so it can stream historical versions for a specific pipeline, which makes more sense in `inspect_pipeline()`.

Context: https://pachyderm.slack.com/archives/C0163RX3YQZ/p1629220407011900?thread_ts=1629219945.011600&cid=C0163RX3YQZ

### DIFF
--- a/src/python_pachyderm/mixin/pps.py
+++ b/src/python_pachyderm/mixin/pps.py
@@ -311,32 +311,51 @@ class PPSMixin:
         """
         return self._req(Service.PPS, "CreatePipeline", req=req)
 
-    def inspect_pipeline(self, pipeline_name, details=None):
+    def inspect_pipeline(self, pipeline_name, history=None, details=None):
         """
-        Inspects a pipeline. Returns a `PipelineInfo` object.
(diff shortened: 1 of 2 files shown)


msteffen (approved):
LGTM

PR #328, opened 2021-08-20
Remove returns from methods whose RPC calls return a negligible protobuf

smalyala:
Negligible protobufs being either Empty protobuf or one of our own protobufs that have no fields.


msteffen (approved):
LGTM

PR #345, opened 2021-09-02
Set up read the docs

### LINKED ISSUE #338: Move from pdoc to ReadTheDocs/sphinx (very low priority for release)
Discussed in 8/18/2021 python-pachyderm product meeting. We currently serve pdoc-generated docs via github pages, but that doesn't give us a way to serve docs for multiple versions of our library. By switching to RTD, we could serve docs from multiple branches and keep things up to date.

smalyala:
- Add read the doc config files
- Convert docstrings for client.py and util.py to numpydoc
- Minor documentation improvements
- Fix a test

Closes #338


src/python_pachyderm/mixin/pfs.py
```diff
@@ -326,17 +328,17 @@ def inspect_commit(
         commit : Union[str, tuple, dict, Commit, pfs_proto.Commit]
             The commit to inspect. Can either be a commit ID or a commit object
             that represents a subcommit (commit at the repo-level).
-        commit_state : pfs_proto.CommitState, optional
+        commit_state : {pfs_proto.CommitState.STARTED, pfs_proto.CommitState.READY, pfs_proto.CommitState.FINISHING, pfs_proto.CommitState.FINISHED}, optional
```
albscui, line 331: Does the `{a, b, c}` syntax help the docs visualize this better?

src/python_pachyderm/mixin/pfs.py
```diff
@@ -1058,7 +1058,9 @@ def diff_file(
 
 
 class ModifyFileClient:
-    """`ModifyFileClient` puts or deletes PFS files atomically."""
+    """``ModifyFileClient`` puts or deletes PFS files atomically. Replaces
+    ``PutFileClient`` from python_pachyderm 6.x.
```
albscui, line 1062: +1, this is very informative.

src/python_pachyderm/client.py
```diff
@@ -198,14 +231,18 @@ def new_from_pachd_address(
...
     @classmethod
-    def new_from_config(cls, config_file):
-        """
-        Creates a Pachyderm client from a config file-like object.
-
-        Params:
-
-        * `config_file`: A file-like object containing the config
-        json file.
+    def new_from_config(cls, config_file: StringIO) -> "Client":
```
albscui, line 234: I think instead of `io.StringIO`, we should use one of the concrete types from the `typing` module. In this case `typing.IO` or `typing.TextIO` makes sense.

src/python_pachyderm/client.py
```diff
@@ -343,6 +375,7 @@ def _parse_config(config):
         host = u.hostname
         port = u.port
         tls = u.scheme == "grpcs" or u.scheme == "https"
+        root_certs = bytes(root_certs, "utf-8") if root_certs is not None else None
```
albscui, line 378: Interesting, is bytes a more robust way for passing the root_certs to gRPC? Just curious about why this change.

albscui (approved):
LGTM! Just some minor comments and nits.

src/python_pachyderm/client.py
```diff
@@ -1,8 +1,10 @@
+from io import StringIO
 import os
 import json
 from pathlib import Path
 from urllib.parse import urlparse
 
+
```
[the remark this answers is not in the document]
smalyala: good catch

docs/requirements.txt
```diff
@@ -0,0 +1,4 @@
+numpydoc
```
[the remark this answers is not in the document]
smalyala: yea this is. one task I have in the queue is to migrate how we're managing dependencies to `poetry`. We can address this requirements.txt situation along with that one :)

src/python_pachyderm/mixin/pfs.py
smalyala: It makes it clear what the possible values are. For other docs, when I see args that can be a limited number of values, the style is typically to list the possible values in a set like so.

src/python_pachyderm/client.py
smalyala: updated

src/python_pachyderm/client.py
smalyala: root_certs is actually supposed to be bytes. It was incorrectly being returned from this helper function as a string.

PR #353, opened 2021-10-13
Mount

smalyala:
- Adds mounting to experimental mixins
- Adds docstrings; will update docs in a later PR where I consolidate some experimental client differences between mount and betterproto
- Fixed CircleCI debug step
- Removed redundant CircleCI script


src/python_pachyderm/experimental/mixin/pfs.py
```diff
@@ -1254,6 +1257,71 @@ def path_exists(
...
+            stderr=subprocess.STDOUT,
+        )
+
+        cmd = ["pachctl", "mount", mount_dir]
+        for r in repos:
+            cmd.append("-r")
+            cmd.append(r)
+
+        subprocess.Popen(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT)
+        time.sleep(0.25)
```
msteffen, line 1296: Any chance we could add a todo to replace the sleep with a health check of some kind (e.g. `ls` the mounted directory, and run ListFile, and compare the results to ensure equality)?

msteffen (approved):
Overall LGTM, with one TODO request

src/python_pachyderm/experimental/mixin/pfs.py
smalyala: Added a check. I find that even with the check, it's necessary to add a delay to give time for mount to complete.

src/python_pachyderm/experimental/mixin/pfs.py
```diff
@@ -1254,6 +1258,85 @@ def path_exists(
...
+
+        # Check for successful mount
+        pach_repos = list(self.list_repo())
+        pach_repos = {repo_info.repo.name for repo_info in pach_repos}
+        repo_names = {r.split("@")[0] for r in repos}
```
msteffen, line 1304: - I think it might make sense to do this check in a loop (in case pachctl mount is taking time to come up) and timeout/error only when N tries to read mounted data have failed
- Would it be possible to shorten this code by simplifying the success criteria to just look at empty/non-empty? Something like
```
for i in range(3):
  time.sleep(0.25)  # Give mount time to start
  try:
    next(os.walk(mount_dir))
    return  # Success
  except StopIteration:
    try:
      next(self.list_repo())
      return # Success
    except StopIteration
      continue
raise RuntimeError("pachctl mount failed to expose data after three read attempts (0.75s)")
self.unmount(repos...)
```

src/python_pachyderm/experimental/mixin/pfs.py
msteffen: See comment below re:retries

msteffen (approved):

src/python_pachyderm/experimental/mixin/pfs.py
smalyala: Loop makes sense to me. Are you suggesting empty/non-empty because comparing repo names as I currently have it discovers the same conditions/outcomes?

src/python_pachyderm/experimental/mixin/pfs.py
```diff
@@ -1254,6 +1258,93 @@ def path_exists(
...
+
+        # Ensure mount has finished
+        for _ in range(3):
+            time.sleep(0.25)
+            mounted_repos = next(os.walk(mount_dir))[1]
+
+            if mounted_repos:
+                return
+
+        self.unmount(mount_dir)
```
lukemarsden, line 1320: why does mount() call unmount()?

smalyala: I noticed a small delay between `pachctl mount` and the actual files showing up in the local directory. Thus, I added a check in the `for` loop to make sure the files actually got mounted. In the rare state that the check fails all 3 times, it `pachctl unmount`s to undo the `mount`.

lukemarsden: got it - thanks! i missed the early exit.

lukemarsden (commented):
is mount() blocking?

src/python_pachyderm/experimental/mixin/pfs.py
lukemarsden: would this still work if the repo was empty?

smalyala: yep!

msteffen (approved):
LGTM

PR #360, opened 2021-11-12
Add betterproto to experimental module

smalyala:
Switches the proto module the mixins in the experimental module use from google protobuf to betterproto-generated code. The main focus for review is in `client.py`. This is where the conversion between google protobuf and betterproto objects occurs, sandwiching the gRPC call (which happens in `_Client._req()`).

Additionally:
- Adds a separate folder of tests for the experimental module in `tests/`
- Updates linting to ignore new proto object attributes
- Renames `python_pachyderm.ExperimentalClient()` to `python_pachyderm.experimental.Client()` 
- Replaces pdoc3 references with sphinx references (`contributing.md`, `__init__.py`)
- Adds `pfs.py`, `util.py`, `service.py` to experimental module
- Generates docs for experimental module


smalyala:
In tox tests, one of the spouts test is failing due to the bug @PFedak discovered. Since this PR doesn't depend on spouts, I'm okay with merging this after review.

etc/proto_lint/proto_lint.py
```diff
@@ -428,7 +428,11 @@ def lint_method(mixin_cls, proto_module, grpc_method_name, mixin_method_name):
 
     # find which arguments differ between the python and gRPC implementation
     request_args = set(
-        [n for n in attrs(request_cls) if n not in PROTO_OBJECT_BUILTINS]
+        [
+            n
+            for n in attrs(request_cls)
+            if n not in PROTO_OBJECT_BUILTINS and "FIELD_NUMBER" not in n
```
smalyala, line 434: @bbonenfant

msteffen (commented):
Note to self/initial thought:
- Is there some way we could preview the generated docs? Since that's a big part of what we hope to get out of this change?
- Did we ever figure out why betterproto-generated objects sometimes use ints for enums and other times use python enums?
- I'm now realizing that this will double the amount of garbage generated per request. That seems mostly fine, but could be a problem for PutFile/GetFile. We should have wrappers for those (ModifyFileClient et. al.) but is there some way this could become a problem? Also: confirm that the wrappers use the Google protobuf objects for efficiency.

msteffen:
> Could you please clarify this more?

Basically, I just meant that the wire-level protocol for PutFile previously looked something like:
```
python-pach ---PutFileReq{"/foo.txt", "co.."    }---> pachd
python-pach ---PutFileReq{"",         "..nte.." }---> pachd
python-pach ---PutFileReq{"",         "..nt"    }---> pachd
```
But because, with this change, we typically create two objects for each RPC, it means that putting a large file might now look more like:
```
python-pach ---BetterProtoPutFileReq{"/foo.txt", "co.."    } ---> GooglePutFileReq{"/foo.txt", "co.."    }---> pachd
python-pach ---BetterProtoPutFileReq{"",         "..nte.." } ---> GooglePutFileReq{"",         "..nte.." }---> pachd
python-pach ---BetterProtoPutFileReq{"",         "..nt"    } ---> GooglePutFileReq{"",         "..nt"    }---> pachd
```

Basically, with PutFile in particular, one logical request might consist of a many smaller stream RPC messages. Because, with this change, we now create two protobuf objects for each request, a BetterProto object and a Google protobuf object, we might be generating twice as much memory garbage in the course of getting a file into pachyderm.

It's probably fine, but I wanted to check that e.g. ModifyFileClient, which just takes a path and a byte source, doesn't bother generating BetterProto objects behind its API and converting them to Google protobuf objects, and instead just sends Google protobuf objects directly, that's all.

msteffen:
> In tox tests, one of the spouts test is failing due to the bug @PFedak discovered. Since this PR doesn't depend on spouts, I'm okay with merging this after review.

For posterity/my own sanity, the issue is this one:
Original question in #eng: https://pachyderm.slack.com/archives/CDWDPKPMY/p1637098926137000
Further discussion in #pach-core: https://pachyderm.slack.com/archives/C02C8TUF493/p1637107996084000
TODO: Link to core pach PR (if any)

tests/experimental/util.py
```diff
@@ -0,0 +1,89 @@
...
+    return "{}{}-{}".format(prefix, test_name, suffix)
+
+
+def create_test_repo(client, test_name, prefix=None, suffix=None):
+    repo_name = test_repo_name(test_name, prefix=prefix, suffix=suffix)
+    client.create_repo(repo_name, "python_pachyderm test repo for {}".format(test_name))
+    return repo_name
+
+
+def create_test_pipeline(client: python_pachyderm.Client, test_name):
```
msteffen: Do you happen to know where this gets used? Is this just carried over from the non-experimental client?

src/python_pachyderm/experimental/mixin/pfs.py
```diff
@@ -4,15 +4,21 @@
...
 import subprocess
 from pathlib import Path
+from contextlib import contextmanager
+from typing import Iterator, Union, List, BinaryIO
 
-from python_pachyderm.pfs import commit_from, Commit, uuid_re
-from python_pachyderm.proto.v2 import pfs
-from python_pachyderm.service import pfs_proto, Service
+from python_pachyderm.experimental.pfs import commit_from, Commit, uuid_re
+from python_pachyderm.service import Service  # , pfs_proto
```
msteffen: Clarifying my [above comment](https://github.com/pachyderm/python-pachyderm/pull/360#issuecomment-988392250) and pinning this for posterity: we could, as a later optimization, import `python_pachyderm.service.pfs_proto` and then change the parts of `ModifyFileClient`'s API that take a URL, byte iterator, or anything else that isn't a protobuf so that they use the non-experimental client's API. Thus instead of:
```
        >>> with client.modify_file_client(c) as mfc:
        >>>     mfc.put_file_bytes(
        ...         "/new_file.txt",
        ...         data)
```
becoming `send(bp_to_pb(experimental.proto.PutFileRequest("/new_file.txt", data)))`, it could be `send(proto.PutFileRequest("/new_file.txt", data))` and avoid the `bp_to_pb` round trip.

msteffen (approved):
Sorry again that this took so long, and thanks for being so patient! Now that I've finally gotten through the review, I see how much tedious rewiring went into making this work. I'm really glad we got this experiment done; I still really think this is something we need to try. Really awesome work.

LGTM

tests/experimental/util.py
smalyala: There are 3 references to `create_test_pipeline()` in `tests/experimental/test_pps.py`.

src/python_pachyderm/experimental/mixin/pfs.py
smalyala: I can add this change to make the `ModifyFile` rpc route not use any betterproto objects.

msteffen (approved):
The new ModifyFileClient LGTM! Nice job keeping the change small!

PR #361, opened 2021-11-16
checking for diff in config file

smalyala:
This PR is being used to investigate why removing the config file in CircleCI allows pachctl commands to not hang, allowing the kube_debug script to run properly.


smalyala:
After each CircleCI step, the shell exists. As a result, `pachctl port-forward` is killed after the `tox` step. In the `kube_debug` step, the client attempts to hit the localhost port which isn't valid, causing the hang. Matthew has a fix that involves adding `nohup` to keep that port-forward process running even after the shell exits. He will also open a PR that adds the PID of `pachctl port-forward` to the pachctl config file to use as reference to see if the process is still running.

PR #363, opened 2021-11-18
Update 7.0.x to use pach 2.0.2 protos

msteffen:


smalyala (approved):
LGTM. Always nice when the protos don't change :)

PR #365, opened 2021-11-18
Revert "Update 7.0.x to use pach 2.0.2 protos"

msteffen:
Reverts pachyderm/python-pachyderm#363

@smalyala and I decided that these changes should go out in a minor release, rather than a patch release. This PR removes them from the `v7.0.x` branch, which should only contain patch-release-level changes to 7.0.0.


msteffen:
I reverted the wrong PR :facepalm: . Deleting this PR

PR #367, opened 2021-11-18
Update docs deps in setup.py

msteffen:


setup.py
```diff
@@ -57,16 +57,23 @@
...
             "pytest>=5.3.4",
             "pytest-mock>=3.6.1",
             "pytest-timeout>=1.4.2",
             "tox>=3.23.1",
+            # docs
             "pdoc3>=0.9.2",
-            "certifi>=2019.11.28",
             "numpydoc>=1.1.0",
+            "Sphinx==4.3.0",
+            "sphinx-rtd-theme==1.0.0",
```
smalyala, line 75: Any reason for fixing these versions?

smalyala (approved):

setup.py
msteffen: Just so that the output docs are stable (i.e. our machines build different docs from the same git commit because we have different versions of these installed).

We used to have this problem in core pach where we weren't careful about this, and then people were constantly panicking that e.g. "we accidentally forgot to build the protos in our last release, because when you run `make proto` in the tagged commit, the generated code is updated" (but actually it was just that the person who did the release had a different version of `protoc` installed). If the versions are fixed, then the output docs should only change when we change the docstrings or change the sphinx et. al. versions.

smalyala: Got it. So it's more to make sure that if the docs change, we know for a fact the change's source is the docstrings that were updated. Not that the sphinx version changed and creates the docs slightly differently.

PR #368, opened 2021-11-18
Update docs build process and rebuild docs

msteffen:
- Update the docs deps in setup.py  (patching #367 into 7.0.x)
- Update `make docs` to use the new docs build process
- Run 'make docs' to rebuild the docs using the fixed version of sphinx et. al.


smalyala:
LGTM

PR #330, opened 2021-08-20
Add release instructions

msteffen:

PR #372, opened 2021-11-19
Update release instructions

msteffen:
- This is essentially #330 for the `v7.0.x` branch
- Also add `twine` (release tool) to the DEV deps in `setup.py`


smalyala (approved):
👍

Issue #280, opened 2021-05-25
Change proto package auth from `albert.cui@pachyderm.io` to `python-pachyderm@pachyderm.io`


msteffen:
When @albscui created the initial python-pachyderm proto package, he set himself as the package author for expediency. However, to make sure that future contributors to the package have access to bugs/questions/whatever, we should instead attach the package to a flexible email alias

albscui:
For context: https://github.com/pachyderm/pachyderm/pull/6182

PR #375, opened 2021-11-30
[INT-442] Poetry

bbonenfant:
Switch from `setup.py` to `poetry`/`pyproject.toml`

This changes the following things:
* No more need for a `setup.py` file, so it gets deleted. Building the wheel is done through `poetry build` and the the dependencies are tracked with `pyproject.toml` and `poetry.lock`
* There's also no need for the `MANIFEST.in` file as `poetry` does all of that in the `pyproject.toml` file. There should probably be a review of this to ensure that the proper things get included in the package that were previously specified in the manifest.
* `poetry` installs the virtual environment at `.venv` instead of `venv`. No way I know around this unfortunately.
* I switched how the `__version__` value gets specified. The modern way to do this is to look at the metadata of the installed package -- this functionality is built into the language in python >=3.8, but there is backport (`importlib-metadata`) for older versions.
    * On this point, `poetry` _must_ have the version specified in the `pyproject.toml` so there should probably be a discussion on how this meshes with the`version.json` file.
* Had to make some tweaks to get CI working (but surprisingly fewer tweaks than I was expecting)
* I changed the `author` of the package on whim to `Pachyderm <python-pachyderm@pachyderm.io>`, but I can revert to JD if that was a mistake.

Follow ons:
* There should be a review and sanitization of the dependency pinning. In general testing dependencies should be pinned and the `^` dependency selector should be used over `>/=>` for package dependencies.


contributing.md
```diff
@@ -2,11 +2,18 @@
...
 
+
+Once both `poetry` is installed, you can use `poetry` to create your
+  virtual environment for this project. The following command
+  (run at the root of this project) will create a virtual environment
+  within the `.venv` directory:
 ```bash
-python3 -m venv venv
-source venv/bin/activate
+poetry install
```
albscui, line 15: Does this install the dev dependencies as well?

albscui (approved):
LGTM!

Although I don't know poetry that well, this looks really useful!

contributing.md
bbonenfant: Yep! When run like this on the repo `poetry` will install all dependencies, including dev dependencies.

smalyala:
Just to confirm, if we want to bump the python-pachyderm version, we need to update it in `pyproject.toml` and `version.json`. As of right now, it seems like the python-pachyderm version in `version.json` is only being used in setup.py (and in `test_version.py`). Correct me if I'm wrong, but we can effectively remove the python-pachyderm line from `version.json` and instead use the version number from `pyproject.toml` for version tests right?

pyproject.toml
```diff
@@ -1,7 +1,55 @@
+[tool.poetry]
+name = "python-pachyderm"
+version = "7.2.0"
+description = "Python Pachyderm Client"
+authors = ["Pachyderm <python-pachyderm@pachyderm.io>"]
```
smalyala, line 5: Let's use the email `integrations@pachyderm.io` instead. @albscui thoughts?

src/python_pachyderm/__init__.py
```diff
@@ -15,9 +15,16 @@
 from grpc import RpcError
 from .experimental import ExperimentalClient
 
+# Python version compatibility.
+try:
+    # >= 3.8
+    import importlib.metadata as metadata
```
smalyala, line 21: Does this read directly from `pyproject.toml`?

pyproject.toml
bbonenfant: I grabbed this from [this issue](https://github.com/pachyderm/python-pachyderm/issues/280) on a whim. Happy to change it to the correct one

src/python_pachyderm/__init__.py
bbonenfant: Nope, packages keep their official version in an associated `METADATA` file that these modules attempt to read (this is how pip can report the version of installed packages without a lock file).

poetry.toml
```diff
@@ -0,0 +1,2 @@
+[virtualenvs]
+in-project = true
```
smalyala, line 2: What does this file do?

bbonenfant: `poetry` by default tries to create all you virtual environments in your home directory, this just tells poetry to create the virtual environment in the project

smalyala: Neat!

smalyala (approved):
LGTM, just one question up above

pyproject.toml
smalyala: Great! This integrations email is the one we're using for managing all integrations-related projects moving forward.

PR #376, opened 2021-12-01
Test publish

chainlink:
This adds publish to testpypi on tag push

To test tag push, we need to push this code to master

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,7 +47,17 @@ jobs:
             echo -e "password = $PYPI_PASSWORD" >> ~/.pypirc
       - run:
           name: Install deps
-          command: pip install twine
+          command: |
+            pip install twine
+            sudo apt install moreutils
+            wget https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 && sudo mkdir -p /home/circleci/bin && mv jq-linux64 /home/circleci/bin/jq
+      - run:
+          name: Compute Version
+          command: |
+            VERSION=`jq -r '."python-pachyderm"' version.json`
+            SHA=`git rev-parse --short HEAD`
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -34,6 +34,30 @@ jobs:
...
+  test-publish:
+    docker:
+      - image: circleci/python:3.7
+    steps:
+      - checkout
+      - run:
+          name: Init .pypirc
+          command: |
+            echo -e "[testpypi]" >> ~/.pypirc
+            echo -e "username = $PYPI_USERNAME" >> ~/.pypirc
```
albscui, line 46:
```suggestion
            echo -e "username = $TEST_PYPI_USERNAME" >> ~/.pypirc
```

.circleci/config.yml
```diff
@@ -34,6 +34,30 @@ jobs:
...
+            echo -e "password = $PYPI_PASSWORD" >> ~/.pypirc
+      - run:
+          name: Install deps
+          command: |
+            pip install twine
```
albscui, line 51:
```suggestion
            python3 -m pip install twine
```

albscui (approved):
Some small suggestions, but LGTM!

.circleci/config.yml
albscui: Although the username would most likely always be `__token__`. I'm assuming eventually we would have two sets of env variables in Circle. `TEST_PYPI_PASSWORD` and `PYPI_PASSWORD`

chainlink: yes, these are set

albscui: I see, my assumption was that env variables are set globally for a project. So if we have two jobs one for testpypi and another for real pypi, then wouldn't we need both `TEST_PYPI_PASSWORD` and `PYPI_PASSWORD`?

PR #381, opened 2021-12-04
[INT-352] Update Client.delete_all

bbonenfant:
This synchronizes what `Client.delete_all` does with what `pachctl delete all` does [here](https://github.com/pachyderm/pachyderm/blob/28e295e6065c2adb093467c676232d87a93b497b/src/client/client.go#L692).

This also adds a new error for when grpc returns an error because `Auth` isn't activated.

I pulled in #379 to add the changes there but that branch isn't up to date with `master` so in order to reduce some noise I have the destination branch of this set to `master` for now.


albscui (approved):
Awesome tests! LGTM!

src/python_pachyderm/errors.py
```diff
@@ -0,0 +1,17 @@
+""" Errors that can be raised by this library. """
+from typing import Union
+
+from grpc import RpcError
+from grpc._channel import _InactiveRpcError
+
+
+class AuthServiceNotActivated(ConnectionError):
```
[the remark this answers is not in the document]
bbonenfant: That's what I originally had, but it's used in both `mixin/auth.py` and `mixin/identity.py` and it felt strange to have the mixins be importing from each other. I could move this file to the `mixin` directory if you'd like that, or if you have any other ideas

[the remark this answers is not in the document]
albscui: Ah I see your point. I think what you have here makes sense! One way to think about this is "how would you expose this to the library's user?".

Right now our user would use it like

```
import python_pachyderm

try:
    some code...
except python_pachyderm.errors.AuthServiceNotActivated:
    handle exception
```

Basically, just try to make it intuitive for the user to handle these custom exceptions.

src/python_pachyderm/mixin/auth.py
```diff
@@ -25,8 +29,15 @@ def activate_auth(self, root_token: str = None) -> str:
     def deactivate_auth(self) -> None:
         """Deactivates auth, removing all ACLs, tokens, and admins from the
         Pachyderm cluster and making all data publicly accessible.
+
+        Raises
```
smalyala, line 33: Can we add this `Raises` section to the other 3 mixin method docstrings too?

src/python_pachyderm/client.py
```diff
@@ -482,6 +483,19 @@ def delete_all(self) -> None:
...
+        except AuthServiceNotActivated:
+            pass
+
+        # Try deactivating auth if activated.
+        try:
+            self.deactivate_auth()
+        except AuthServiceNotActivated:
+            pass
+
+        self.delete_all_license()
```
smalyala, line 498: Does `delete_all_license()` also need to be put in a `try-except`? https://github.com/pachyderm/pachyderm/blob/28e295e6065c2adb093467c676232d87a93b497b/src/client/client.go#L705

smalyala (approved):
LGTM, this is a really good change for parity with core pach behavior. Left a couple comments.

src/python_pachyderm/client.py
bbonenfant: Ah I never hit an error when testing which is why I didn't add this, but if core does it than I'll add it to the license mixins

PR #383, opened 2021-12-10
[WIP] Refactor gRPC method infrastructure

bbonenfant:
Here's a preliminary refactor of the `AdminMixin` and `AuthMixin` classes where requests are no longer routed through the `Client` class. Doing so this way I believe increases readability and would drastically reduce the linting we do for these mixins. If this was implemented in all the mixin classes then I believe that the `self._req` method would be unnecessary as well as `service.py`. 

Currently the only unpleasant part of this setup is that it requires each method to pass the metadata object in, but I believe even that can be alleviated with a gRPC interceptor. Using something like [this](https://github.com/d5h-foss/grpc-interceptor) we could attach a client interceptor to the `Client._channel` object which intercepts all the outgoing grpc buffers and appends the required metadata. If that turns out to be true than this could be really ergonomic to use!

### DIFF
--- a/src/python_pachyderm/client.py
+++ b/src/python_pachyderm/client.py
@@ -1,8 +1,10 @@
 import os
 import json
 from pathlib import Path
-from typing import TextIO
+from typing import List, Optional, TextIO, Tuple
 from urllib.parse import urlparse
+
+import grpc
 
 from .mixin.admin import AdminMixin
 from .mixin.auth import AuthMixin
@@ -15,7 +17,7 @@ from .mixin.pfs import PFSMixin
 from .mixin.pps import PPSMixin
 from .mixin.transaction import TransactionMixin
 from .mixin.version import VersionMixin
-from .service import Service
+from .service import Service, GRPC_CHANNEL_OPTIONS
 
 
 class ConfigError(Exception):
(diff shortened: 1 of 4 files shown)


smalyala:
Code changes make sense to me. Definitely increases readability, I'm a fan of that. How would this simplify linting?

bbonenfant:
> Code changes make sense to me. Definitely increases readability, I'm a fan of that. How would this simplify linting?

Well the current proto linting appears to check that all of the RPCs are implemented within their respective mixin class (except non-supported RPCs), and that the arguments to the python methods map as expected to the RPC request parameters. I would argue if the objective of this library is to provide a more "pythonic" interface to this gRPC API, then we only need to be checking the gRPC side of things as opposed to the python side. 

For example, we probably don't need to be checking if there exists an `inspect_branch` method on the `PFSMixin` class that has `repo_name` and `branch_name` as arguments, we just need to ensure that a method of this class calls the `InspectBranch` RPC. From there we can make sure the proper gRPC Message parameters are being set, but even that is probably better covered with unit/integration tests to ensure the RPC calls themselves perform the proper actions and/or verify this with static type-checking. Code review would ensure that the methods and their arguments have logical names.

How this linting is implemented is up to interpretation. The simplest would probably be to just regex the proto source files for the RPC names and ensure the respective `mixin/<module>.py` file calls these methods on the APIStub objects (they should all be named the same). Or, a more controlled approach could be to mock all the APIStub objects and ensure the RPC calls get expected Message objects.

bbonenfant:
Additionally, [this link](https://github.com/pachyderm/python-pachyderm/compare/bonenfant/grpc_request_refactor...bonenfant/grpc-request-refactor-interceptor) shows how the gRPC interceptor I referenced in the top comment would fit into this (still a WIP though)

smalyala (approved):
I'm good with this change in how we call the RPCs. @msteffen Any thoughts on this before we adopt this shift?

PR #389, opened 2022-01-28
Fix for betterproto initialisms

bbonenfant:
Betterproto's code generation seems confused by IDP and OIDC in class names, and changing them to Idp and Oidc respectively fixed things


smalyala (approved):
Nice!

PR #390, opened 2022-01-31
Refactor stable API to not use `client._req`

bbonenfant:
Decided to split the task into refactoring stable and refactoring the experimental API because it's diverged quite a bit. This is also the reason why I'm making a new PR as opposed to continuing #383 

JIRA issues: [INT-492]

[INT-492]: https://pachyderm.atlassian.net/browse/INT-492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/python_pachyderm/mixin/auth.py
```diff
@@ -177,7 +186,8 @@ def authenticate_oidc(self, oidc_state: str) -> str:
         str
             A token that can be used for making authenticate requests.
         """
-        return self._req(Service.AUTH, "Authenticate", oidc_state=oidc_state).pach_token
+        message = auth_pb2.AuthenticateRequest(oidc_state=oidc_state)
+        return self.__stub.Authorize(message).pach_token
```
smalyala, line 190: This should be `self.__stub.Authenticate()`. Might be good to get coverage on these methods in the tests.

src/python_pachyderm/mixin/auth.py
```diff
@@ -194,13 +204,14 @@ def authenticate_id_token(self, id_token: str) -> str:
         str
             A token that can be used for making authenticate requests.
         """
-        return self._req(Service.AUTH, "Authenticate", id_token=id_token).pach_token
+        message = auth_pb2.AuthenticateRequest(id_token=id_token)
+        return self.__stub.Authorize(message).pach_token
```
smalyala, line 208: This should be `self.__stub.Authenticate()`.

src/python_pachyderm/mixin/debug.py
```diff
@@ -1,13 +1,22 @@
...
+from python_pachyderm.proto.v2.debug import debug_pb2, debug_pb2_grpc
+
 
 class DebugMixin:
     """A mixin for debug-related functionality."""
 
+    _channel: grpc.Channel
+
+    def __init__(self):
+        self.__stub = debug_pb2_grpc.DebugStub(self._channel)
```
smalyala, line 15: It's funny to me that proto decides to spit out `DebugStub` rather than `ApiStub` here.

smalyala:
Check out `util.py`. That has some `pps_proto` references.

smalyala:
What are your thoughts on also changing module names in the docstrings (i.e. "Returns `pfs_proto.Repo` -> `pb2_proto.Repo`")?

src/python_pachyderm/mixin/pfs.py
```diff
@@ -1047,18 +1038,18 @@ def list_file(
         --------
         >>> files = list(client.list_file(("foo", "master"), "/dir/subdir/"))
         """
-        return self._req(
-            Service.PFS,
-            "ListFile",
-            file=pfs_proto.File(commit=commit_from(commit), path=path, datum=datum),
+        message = pfs_pb2.ListFileRequest(
+            details=details,
```
smalyala, line 1042: No harm in adding it back in but fyi `details` isn't actually used

src/python_pachyderm/mixin/auth.py
bbonenfant: Good catch, I'd definitely like to increase the coverage here, but I'm not exactly sure of a great way to test it. any ideas?

smalyala (approved):
Sorry, lost track of this PR. LGTM!

PR #391, opened 2022-02-02
[Bugfix] Cannot get large files from PFS

bbonenfant:
This fixes an issue where a users could not retrieve more than 20MiB of a file using `client.get_file`. This also creates a working implementation for `client.get_file_tar`.

### DIFF
--- a/src/python_pachyderm/mixin/pfs.py
+++ b/src/python_pachyderm/mixin/pfs.py
@@ -5,56 +5,26 @@ import tarfile
 from contextlib import contextmanager
 from typing import Iterator, Union, List, BinaryIO
 
+import grpc
+from betterproto import BytesValue
+
 from python_pachyderm.pfs import commit_from, uuid_re, SubcommitType
 from python_pachyderm.service import pfs_proto, Service
 from google.protobuf import empty_pb2, wrappers_pb2
 
-
 BUFFER_SIZE = 19 * 1024 * 1024
 
 
-class FileTarstream:
-    """Implements a file-like interface over a GRPC byte stream,
-    so we can use tarfile to decode the file contents.
-    """
-
-    def __init__(self, res):
-        self.res = res
-        self.buf = []
-
-    def __next__(self):
-        return next(self.res).value
-
-    def close(self):
-        self.res.cancel()
-
-    def read(self, size=-1):
-        if self.res.cancelled():
-            return b""
-
-        buf = []
-        remaining = size if size >= 0 else 2 ** 32
-
-        if self.buf:
-            buf.append(self.buf[:remaining])
-            self.buf = self.buf[remaining:]
-            remaining -= len(buf[-1])
-
-        try:
-            while remaining > 0:
-                b = next(self)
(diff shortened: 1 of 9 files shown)


tests/experimental/test_pfs.py
```diff
@@ -773,3 +734,138 @@ def test_mount():
...
+        assert streamed_data[:10] == data[:10]
+        assert streamed_data[-10:] == data[-10:]
+        assert len(streamed_data) == len(data)
+
+    @staticmethod
+    def test_buffered_data(client: Client, repo: str):
+        """Test that data gets buffered as expected."""
+        # Arrange
+        stream_items = [b"a", b"bc", b"def", b"ghij", b"klmno"]
+        stream = (BytesValue(value=item) for item in stream_items)
```
smalyala, line 785: This parenthesis syntax works?

bbonenfant: Yep, it's the comprehension-syntax for iterators, basically just a lazy list comprehension

smalyala: Sick. I didn't know generators had their own comprehension too.

src/python_pachyderm/mixin/pfs.py
```diff
@@ -111,11 +71,28 @@ def read(self, size: int = -1) -> bytes:
         bytes
             Content from the stream.
         """
-        return self._file.read(size)
+        try:
+            if size == -1:
+                # Consume the entire stream.
+                for message in self._stream:
```
smalyala, line 77: Assuming an infinite stream (or absurdly large stream), should we cap how much we read in before returning so this isn't stuck in an infinite loop?

bbonenfant: That's a good question, I would say that it's fine and I'd rather just be consistent with how `.read()` works with standard python file objects.

Also fun fact, I tried this to see if python would let me do it and I just hard crashed my computer running out of memory:
```python
with open("/dev/urandom") as evil:
    evil.read()
```

src/python_pachyderm/mixin/pfs.py
```diff
@@ -1,60 +1,31 @@
...
-                else:
-                    buf.append(b)
-
-                remaining -= len(buf[-1])
-        except StopIteration:
-            pass
-
-        return b"".join(buf)
+class PfsTarFile(tarfile.TarFile):
+    def __iter__(self):
```
smalyala, line 20: Is this called during `extractall()`?

smalyala (approved):
Left few comments. Really good work with the tests!

src/python_pachyderm/mixin/pfs.py
bbonenfant: Yeah, the inner working of `tarfile.TarFile` is kind of strange, it iterates over itself to get the individual member of the TAR. This seemed like the easiest way to insert custom code into that

smalyala: Solid

PR #392, opened 2022-02-03
Prep for v7.1.1 release

bbonenfant:


tests/experimental/test_pps.py
```diff
@@ -89,11 +89,11 @@ def test_stop_job():
...
-    # This is necessary because `StopJob` does not wait for the job to be
-    # killed before returning a result.
-    # TODO: remove once this is fixed:
-    # https://github.com/pachyderm/pachyderm/issues/3856
-    time.sleep(1)
+    # # This is necessary because `StopJob` does not wait for the job to be
+    # # killed before returning a result.
+    # # TODO: remove once this is fixed:
+    # # https://github.com/pachyderm/pachyderm/issues/3856
+    # time.sleep(1)
```
smalyala, line 96: Why comment this out?

smalyala:
If we're doing this as a patch release, we should cherry pick the bugfix commit into `7.1.x` and release off that minor version branch (https://github.com/pachyderm/python-pachyderm/blob/master/contributing.md#update-versions-for-next-release). I can also see a case for doing a `7.2.0` minor release off `master` since we added betterproto and the more complete `Client.delete_all()`.

Also, probably won't change anything but can you bump pach version up to 2.0.7 and rerun proto gen?

tests/experimental/test_pps.py
bbonenfant: Whoops, didn't mean to.

bbonenfant:
> If we're doing this as a patch release, we should cherry pick the bugfix commit into `7.1.x` and release off that minor version branch (https://github.com/pachyderm/python-pachyderm/blob/master/contributing.md#update-versions-for-next-release). I can also see a case for instead doing a `7.2.0` minor release off `master` since we added betterproto and the more complete `Client.delete_all()`.
> 
> Also, probably won't change anything but can you bump pach version up to 2.0.7 and rerun proto gen?

Oh I didn't realize there was a `7.1.x` branch. In that case yeah I'll make this `7.2.0` and cherry pick the bug fix into `7.1.1`. As for the pachyderm version, I put it at `2.0.5` because it's the first version to have the spout fix, but I guess I'm not certain about whether we should be pinning the most recent version or the oldest version that python_pachyderm supports.

smalyala:
> > If we're doing this as a patch release, we should cherry pick the bugfix commit into `7.1.x` and release off that minor version branch (https://github.com/pachyderm/python-pachyderm/blob/master/contributing.md#update-versions-for-next-release). I can also see a case for instead doing a `7.2.0` minor release off `master` since we added betterproto and the more complete `Client.delete_all()`.
> > Also, probably won't change anything but can you bump pach version up to 2.0.7 and rerun proto gen?
> 
> Oh I didn't realize there was a `7.1.x` branch. In that case yeah I'll make this `7.2.0` and cherry pick the bug fix into `7.1.1`. As for the pachyderm version, I put it at `2.0.5` because it's the first version to have the spout fix, but I guess I'm not certain about whether we should be pinning the most recent version or the oldest version that python_pachyderm supports.

That's a good point. What we can do is regen the protos with `2.0.7` and see if they change. Based on that we can see if there are any changes we need to do (I'm guessing the proto changes will be trivial).

bbonenfant:
> > > If we're doing this as a patch release, we should cherry pick the bugfix commit into `7.1.x` and release off that minor version branch (https://github.com/pachyderm/python-pachyderm/blob/master/contributing.md#update-versions-for-next-release). I can also see a case for instead doing a `7.2.0` minor release off `master` since we added betterproto and the more complete `Client.delete_all()`.
> > > Also, probably won't change anything but can you bump pach version up to 2.0.7 and rerun proto gen?
> > 
> > 
> > Oh I didn't realize there was a `7.1.x` branch. In that case yeah I'll make this `7.2.0` and cherry pick the bug fix into `7.1.1`. As for the pachyderm version, I put it at `2.0.5` because it's the first version to have the spout fix, but I guess I'm not certain about whether we should be pinning the most recent version or the oldest version that python_pachyderm supports.
> 
> That's a good point. What we can do is regen the protos with `2.0.7` and see if they change. Based on that we can see if there are any changes we need to do (I'm guessing the proto changes will be trivial).

Oh yeah, just checked and the only proto changes are in `proxy.proto` so I think we are ok staying with `2.0.5`

smalyala (approved):
LGTM

smalyala (approved):

PR #398, opened 2022-04-26
Add check_pachctl utility method

bbonenfant:
Add check that `pachctl` is present on the system before calling mount operations.

This also exposes a `check_pachctl` method within `python_pachyderm.experimental.util`.

### DIFF
--- a/src/python_pachyderm/experimental/util.py
+++ b/src/python_pachyderm/experimental/util.py
@@ -1,15 +1,20 @@
 import os
-from typing import Union
+from shutil import which
+from subprocess import run
+from typing import Union, TYPE_CHECKING
 
-from python_pachyderm.experimental import Client
-from python_pachyderm.experimental.pfs import Commit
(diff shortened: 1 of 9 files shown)


src/python_pachyderm/experimental/util.py
```diff
@@ -143,3 +148,20 @@ def parse_dict_pipeline_spec(d: dict) -> pps_proto.CreatePipelineRequest:
...
+    ------
+    FileNotFoundError
+        No pachctl binary present
+    RuntimeError
+        pachctl binary present but cannot run mount operations
+    """
+    if which("pachctl") is None:
+        raise FileNotFoundError("pachctl")
+    if run(["pachctl", "mount", "-h"], capture_output=True).returncode != 0:
+        raise RuntimeError("incompatible pachctl detected")
```
pleeko, line 167: I like the double check here!

pleeko (commented):
LGTM

PR #399, opened 2022-05-02
release v7.2.0

bbonenfant:
* Better handling of gRPC Connection errors when unable to connect to a pachyderm instance.
* Added checks for pachctl when using mount operations and exposed python_pachyderm.experimental.util.check_pachctl method.


smalyala (approved):
LGTM

PR #400, opened 2022-05-05
Add support for python3.10

bbonenfant:
For some reason the installing python3.10 through `apt` broke CI, so I had to do some research on how on a more sustainable method for testing python on circleci. The major changes were:
* using `pyenv` to install the python version used for testing
* abandon `tox` for CI (it's still a dev dependency and referenced in the contributor readme, but it can be fully removed if you think it no longer is useful)
* rearrange the circleCI config to make it more extensible and easier to perform these updates in the future
* Also had to bump `betterproto` version for python3.10 compatibility.

Let me know what you think

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,63 @@
 version: 2.1
 
+commands:
+  install-python:
+    parameters:
+      version:
+        type: string
+    steps:
+        - run:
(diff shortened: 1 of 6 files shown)


smalyala (approved):
The config rearrangements will be very handy in the future. LGTM!

PR #401, opened 2022-06-01
pin protobuf dependency to >=3.17.1,<4.0.0

bbonenfant:
A non-patch release will need to happen to update to protobuf>=4.21.1 as it causes incompatibility issues with protobufs generated with protobuf version >=3.17.1,<4.0.0.


pleeko (approved):
LGTM

PR #404, opened 2022-06-22
[INT-604] Validation for user-provided root certs

bbonenfant:
Adds a check that the `root_certs` provided by the user has type `bytes` and a light check that it is in the PEM format.


smalyala (approved):
LGTM

Issue #410, opened 2022-09-22
`Client.get_file` silently ignores parameters


defiori:
Starting in ` v7.2.1`, the `URL` and `offset` parameters to [`Client.get_file`](https://github.com/pachyderm/python-pachyderm/blob/v7.2.1/src/python_pachyderm/mixin/pfs.py#L887) are ignored, is this an intentional breaking change?

bbonenfant:
This was not intentional. Thank you for the bug repot, we will get this fixed as soon as possible.

bbonenfant:
This fix was released in version 7.3.3 of python-pachyderm

PR #412, opened 2022-09-23
revert breaking change

bbonenfant:
Fix a breaking change to the `Client.get_file` method which silently ignored the `URL` and `offset` arguments.

closes: #410


smalyala (approved):
LGTM

PR #413, opened 2022-11-09
INT-747 Projects Support

bbonenfant:


bbonenfant:
One last `StopIteration` class of errors I still need to track down, but overall this is ready for review.

Makefile
```diff
@@ -13,7 +13,7 @@ src/python_pachyderm/proto/v2: docker-build-proto
 		rm -rf src/python_pachyderm/experimental/proto/v2
 	cd proto/pachyderm && \
 		git fetch --all && \
-		git checkout v$(PACHYDERM_VERSION)
```
bbonenfant, line 16: This should be reverted before merging

src/python_pachyderm/experimental/mixin/pfs.py
```diff
@@ -361,12 +468,15 @@ def list_commit(
         .. # noqa: W505
         """
         if repo_name is not None:
+            if started_time is not None:
+                started_time = timestamp_pb2.Timestamp.FromDatetime(started_time)
```
smalyala, line 472: Does betterproto generate an equivalent for this?

src/python_pachyderm/experimental/mixin/pfs.py
```diff
@@ -826,6 +981,7 @@ def get_file(
         datum: str = None,
         URL: str = None,
         offset: int = 0,
+        path_range: pfs_proto.PathRange = None,
```
smalyala, line 984: Param missing docstring

src/python_pachyderm/experimental/mixin/pfs.py
```diff
@@ -988,7 +1160,10 @@ def walk_file(
         )
 
     def glob_file(
-        self, commit: Union[tuple, dict, Commit, pfs_proto.Commit], pattern: str
+        self,
+        commit: Union[tuple, dict, Commit, pfs_proto.Commit],
+        pattern: str,
+        path_range: pfs_proto.PathRange = None,
```
smalyala, line 1166: param missing docstring

src/python_pachyderm/experimental/mixin/pps.py
```diff
@@ -82,6 +88,8 @@ def list_job(
         history: int = 0,
         details: bool = False,
         jqFilter: str = None,
+        project_name: str = None,
+        project_filter: List[str] = None,
```
smalyala, line 92: Missing docstring and usage of `project_filter`

src/python_pachyderm/pfs.py
```diff
@@ -36,9 +41,11 @@ def from_pb(commit: pfs_pb2.Commit) -> "Commit":
             branch=commit.branch.name,
             id=commit.id,
             repo_type=commit.branch.repo.type,
+            project=commit.branch.repo.project.name,
         )
 
 
+# TODO: How to reconcile tuple commit format with projects.
```
smalyala, line 48: This is a tough one. I'm leaning towards `(repo, brach-or-commit-id, project, commit-id, repoType)`. We should keep repo and branch-or-commit-id as the first two as they're necessary and sufficient. To me, projects feels like it would be more referenced than commit-id, but again, that's just a guess.

smalyala:
Great work here! This must have been tedious. Left a few comments.

src/python_pachyderm/experimental/mixin/pps.py
bbonenfant: Yeah the "projects_filter" (named "projects" in the proto messages) are a little un-intuitive in my opinion. Any projects that are listed under the "project_filter" are filtered _out_ of the response.

Moreover, having both "project_name" and "project_filter" is really not ideal. We should just be having users pass "Repo" objects that contain the information like project, repo, branch, etc. all bundled together. But that would be a huge change and we are trying to make this as backwards-compatible as possible.

src/python_pachyderm/experimental/mixin/pfs.py
bbonenfant: This was actually meant to be internal. I removed it.

src/python_pachyderm/experimental/mixin/pfs.py
```diff
@@ -416,6 +421,7 @@ def list_commit(
         all: bool = False,
         origin_kind: pfs_proto.OriginKind = pfs_proto.OriginKind.USER,
         started_time: datetime = None,
+        project_name: str = None,
```
smalyala, line 424: Wouldn't the user have to pass a `pfs_proto.Project` object here?

src/python_pachyderm/experimental/mixin/pps.py
```diff
@@ -118,8 +121,19 @@ def list_job(
         jqFilter : str, optional
             A ``jq`` filter that can filter the list of jobs returned, only if
             `pipeline_name` is provided.
-        project_name : str
+        project_name : str, optional
```
smalyala, line 124: Should we name this `pipeline_project_name` to make it clear this arg is to help specify which pipeline we're filtering jobs on?

src/python_pachyderm/experimental/mixin/pps.py
```diff
@@ -604,6 +626,7 @@ def list_pipeline(
         details: bool = False,
         jqFilter: str = None,
         commit_set: pfs_proto.CommitSet = None,
+        projects_filter: List[str] = None,
```
smalyala, line 629: re: `str` vs `pfs_proto.Project` as mentioned above

src/python_pachyderm/mixin/pfs.py
```diff
@@ -480,6 +485,7 @@ def list_commit(
         all: bool = False,
         origin_kind: pfs_pb2.OriginKind = pfs_pb2.OriginKind.USER,
         started_time: datetime = None,
+        project_name: str = None,
```
smalyala, line 488: re: `str` vs `pfs_pb2.Project` as mentioned above

src/python_pachyderm/mixin/pps.py
```diff
@@ -84,6 +84,10 @@ def list_job(
         details: bool = False,
         jqFilter: str = None,
         project_name: str = None,
+        projects_filter: List[str] = None,
```
smalyala, line 87: re: `str` vs `pfs_pb2.Project` as mentioned above

src/python_pachyderm/mixin/pps.py
```diff
@@ -112,8 +116,19 @@ def list_job(
         jqFilter : str, optional
             A ``jq`` filter that can filter the list of jobs returned, only if
             `pipeline_name` is provided.
-        project_name : str
-            The name of the project.
+        project_name : str, optional
```
smalyala, line 119: re: updating this name to make it clear, similar to mentioned above

src/python_pachyderm/mixin/pps.py
```diff
@@ -571,6 +596,7 @@ def list_pipeline(
         details: bool = False,
         jqFilter: str = None,
         commit_set: pfs_pb2.CommitSet = None,
+        projects_filter: List[str] = None,
```
smalyala, line 599: re: `str` vs `pfs_pb2.Project` as mentioned above

smalyala (approved):

src/python_pachyderm/experimental/mixin/pfs.py
bbonenfant: Good catch, we do want them to provide strings, just like we have users give use strings instead of `pfs_proto.Repo` objects, but we need to construct those in the message. Adding tests to catch this in the future. This comment applies to all below as well

src/python_pachyderm/experimental/mixin/pps.py
bbonenfant: I'm going to decide to leave it just to be consistent with all of the other methods that use "project_name", but this does raise a good point that these new apis are getting to be confusing.

PR #415, opened 2023-01-18
Project support bugs - PPS Mixin

bbonenfant:
Several bugs in the PPS Mixin classes regarding specifying a pipeline in relation to a project. Additionally added some more tests for project methods within the PFS mixin. 

Also of note, had to update the `test_pps.py::test_get_pipeline_logs` tests because of an issue where the combination of the project and pipeline names were too long. 

Still looking into the one tests that is failing `experimental/test_pps.py::test_delete_pipeline`. I'm not yet sure why only the experimental test is breaking and why I can't get it to break locally.


smalyala (approved):
LGTM

PR #417, opened 2023-02-01
fix default argument list bug

bbonenfant:
This can cause unexpected issues, as described [here](https://stackoverflow.com/q/1132941)


smalyala (approved):
Ooo this is interesting, never seen this before. One question I have is will that unexpected behavior affect us, since we're not modifying the repos list within the function?

bbonenfant:
Nope, it only becomes a problem once you start modifying the list. That's why this ticket was able to hang around for over a year, haha

PR #433, opened 2023-05-29
Datum Batching CI

chainlink:


.circleci/config.yml
```diff
@@ -94,7 +97,7 @@ jobs:
           command: |
             pachctl port-forward &
             mkdir -p test-results
-            poetry run pytest -v --junitxml=test-results/junit.xml tests
+            DATUM_BATCHING_IMAGE=pachyderm/python-datum-batching-ci:$CIRCLE_SHA1 poetry run pytest -v --junitxml=test-results/junit.xml tests
```
bbonenfant, line 100: Can you store this `DATUM_BATCHING_IMAGE` in the environment at a higher level so that build, publish, and test can use it?

tests/test_worker.py
```diff
@@ -1,11 +1,14 @@
...
+import os
+
 from python_pachyderm import Client
 from python_pachyderm.proto.v2.pps import pps_pb2
 
 from .test_pfs import _client_fixture, _repo_fixture
 
-IMAGE_NAME = "bonenfan5ben/datum_batching:0037pp"  # TODO: Don't do this.
+IMAGE_NAME = os.environ.get("DATUM_BATCHING_IMAGE")
+# IMAGE_NAME = "bonenfan5ben/datum_batching:0037pp"  # TODO: Don't do this.
```
bbonenfant, line 11: I think we can delete this

bbonenfant (approved):
Looks good to me

.circleci/config.yml
chainlink: We can, but just to check my understanding, is the env var used in those other jobs?

PR #441, opened 2023-08-14
Proto regen for 2.7.0

bbonenfant:
The main change is moving to the `/Debug/DumpV2` api. Additionally the `details_json` field for /Pps/CreatePipeline` was reserved in the protos but support has not yet been implemented

### DIFF
--- a/src/python_pachyderm/mixin/debug.py
+++ b/src/python_pachyderm/mixin/debug.py
@@ -1,9 +1,10 @@
-from typing import Iterator
+from typing import Iterator, List
 
 import grpc
 from google.protobuf import duration_pb2
 
 from python_pachyderm.proto.v2.debug import debug_pb2, debug_pb2_grpc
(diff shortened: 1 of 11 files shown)


src/python_pachyderm/mixin/debug.py
```diff
@@ -16,34 +17,80 @@ def __init__(self):
...
+
+        Parameters
+        ----------
+        filters : List[str], optional
+            No supported filters - this argument has no effect.
+
+        Yields
+        -------
+        debug_pb2.DumpV2Request
+            The request to
```
smalyala, line 83: This line might not have saved

smalyala (approved):

src/python_pachyderm/mixin/debug.py
bbonenfant: I guess it's the request to nowhere, haha

